PR #2, opened 2024-12-13
feat: GKR LogUp

zlangley:


crates/stark-backend/src/interaction/stark_log_up.rs
```diff
@@ -65,7 +66,7 @@ where
     type PartialProof = ();
     type ProvingKey = StarkLogUpProvingKey;
```
jonathanpwang, line 67: do we still need this? I'm not sure if philosophically RAP phase should every require more information than the rest of the proving key provides

jonathanpwang: nvm FRI logup does require it in the permutation trace generation. hmm it's kind of weird

zlangley: This is actually only needed for FRI logup. Would love to get rid of it.

PR #13, opened 2024-12-20
[feat] SymbolicConstraints as DAG

nyunyunyunyu:
To avoid exponential blowup in the evaluation of the symbolic syntax tree, we convert it into a DAG with topological sort and evaluate topologically (currently single threaded) to avoid there is not duplicate evaluations. The main note is that `Arc` should be hashed using `ptr::hash`.

This PR will not switch the serialization of vkey to DAG. This will be done in a followup where we switch the main interfaces to use the DAG, and `SymbolicExpression` should only be an internally used struct.

towards INT-2905

### DIFF
--- a/crates/stark-backend/src/air_builders/symbolic/dag.rs
+++ b/crates/stark-backend/src/air_builders/symbolic/dag.rs
@@ -1,0 +1,266 @@
+use std::sync::Arc;
+
+use p3_field::Field;
+use rustc_hash::FxHashMap;
+use serde::{Deserialize, Serialize};
+
+use crate::air_builders::symbolic::{
+    symbolic_expression::SymbolicExpression, symbolic_variable::SymbolicVariable,
+};
+
+/// A node in symbolic expression DAG. Basically replace `Arc`s in `SymbolicExpression` with node IDs.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(bound = "F: Field")]
+enum SymbolicExpressionNode<F> {
+    Variable(SymbolicVariable<F>),
+    IsFirstRow,
+    IsLastRow,
+    IsTransition,
+    Constant(F),
+    Add {
+        x: usize,
+        y: usize,
+        degree_multiple: usize,
+    },
+    Sub {
+        x: usize,
+        y: usize,
+        degree_multiple: usize,
+    },
+    Neg {
+        x: usize,
+        degree_multiple: usize,
+    },
+    Mul {
+        x: usize,
+        y: usize,
+        degree_multiple: usize,
+    },
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(bound = "F: Field")]
+pub struct SymbolicExpressionDag<F> {
+    /// Nodes in **topological** order.
+    expr_by_id: Vec<SymbolicExpressionNode<F>>,
+    /// Node ids of expressions to assert.
+    top_expr_ids: Vec<usize>,
+}
+
+pub(super) fn build_symbolic_expr_dag<F: Field>(
+    exprs: &[SymbolicExpression<F>],
+) -> SymbolicExpressionDag<F> {
+    let mut expr_to_id = FxHashMap::default();
+    let mut id_to_expr = Vec::new();
+    let top_expr_ids = exprs
+        .iter()
+        .map(|expr| topology_sort_symbolic_expr(expr, &mut expr_to_id, &mut id_to_expr))
+        .collect();
+    SymbolicExpressionDag {
+        expr_by_id: id_to_expr,
+        top_expr_ids,
+    }
+}
+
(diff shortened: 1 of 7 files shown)


crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -0,0 +1,266 @@
...
+#[serde(bound = "F: Field")]
+enum SymbolicExpressionNode<F> {
+    Variable(SymbolicVariable<F>),
+    IsFirstRow,
+    IsLastRow,
+    IsTransition,
+    Constant(F),
+    Add {
+        x: usize,
+        y: usize,
```
zlangley, line 22: Maybe `left_id` and `right_id` are better to emphasize that this is not adding `x` and `y` but rather the gates.

crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -0,0 +1,266 @@
...
+        .iter()
+        .map(|expr| topology_sort_symbolic_expr(expr, &mut expr_to_id, &mut id_to_expr))
+        .collect();
+    SymbolicExpressionDag {
+        expr_by_id: id_to_expr,
+        top_expr_ids,
+    }
+}
+
+fn topology_sort_symbolic_expr<'a, F: Field>(
```
zlangley, line 65: `topological_sort`

crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -0,0 +1,266 @@
...
+        SymbolicExpression::IsLastRow => SymbolicExpressionNode::IsLastRow,
+        SymbolicExpression::IsTransition => SymbolicExpressionNode::IsTransition,
+        SymbolicExpression::Constant(cons) => SymbolicExpressionNode::Constant(*cons),
+        SymbolicExpression::Add {
+            x,
+            y,
+            degree_multiple,
+        } => {
+            let x_id = topology_sort_symbolic_expr(x.as_ref(), expr_to_id, id_to_expr);
+            let y_id = topology_sort_symbolic_expr(y.as_ref(), expr_to_id, id_to_expr);
```
zlangley, line 85: you should check that `y_id != x_id`; otherwise you've already done the work.

crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -0,0 +1,266 @@
...
+        .collect();
+    SymbolicExpressionDag {
+        expr_by_id: id_to_expr,
+        top_expr_ids,
+    }
+}
+
+fn topology_sort_symbolic_expr<'a, F: Field>(
+    expr: &'a SymbolicExpression<F>,
+    expr_to_id: &mut FxHashMap<&'a SymbolicExpression<F>, usize>,
```
zlangley, line 67: I think we can remove this. (per slack discussion)

crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -0,0 +1,266 @@
...
+    },
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(bound = "F: Field")]
+pub struct SymbolicExpressionDag<F> {
+    /// Nodes in **topological** order.
+    expr_by_id: Vec<SymbolicExpressionNode<F>>,
+    /// Node ids of expressions to assert.
+    top_expr_ids: Vec<usize>,
```
zlangley, line 47: These are basically output gates right? `output_ids` might be a better name

crates/stark-backend/src/air_builders/symbolic/dag.rs
jonathanpwang: done

jonathanpwang (approved):
Custom hash implementation using `ptr::hash` on `Arc`.
I want to split out `SymbolicInteractions` from the vkey and make `SymbolicExpressionDag` the main struct that is used for verification, but merging just this change to resolve the exponential blowup issue.

crates/stark-backend/src/air_builders/symbolic/symbolic_expression.rs
```diff
@@ -43,6 +47,36 @@ pub enum SymbolicExpression<F> {
     },
 }
 
+impl<F: Field> Hash for SymbolicExpression<F> {
```
zlangley, line 50: Why do we need this? I thought we were just using the raw pointer in the hashmap.

jonathanpwang: This was slightly better because without it, x.clone() * x.clone(), x.clone * x.clone(), record two separate Mul

jonathanpwang: In the unit test, if you do just the pointer hashmap, there are a bunch more nodes

PR #19, opened 2025-01-07
feat: quotient polynomial evaluation from serialized constraints

jonathanpwang:
`ProverConstraintFolder -> ProverConstraintEvaluator` is now no longer an `AirBuilder` and moved into `prover/quotient` since it is only used for internal quotient polynomial evaluation.
- Still follows the general pattern of parallelizing across row chunks, where each row chunk is evaluated using SIMD `PackedVal, PackedChallenge`. Now it does it by evaluating topologically sorted nodes in the constraint DAG.

I did not change the high level `prove` interface: that will have to wait for a future PR: `AnyRap` is still needed for the debug builder (if you want rust backtrace of the exact line where a constraint failed, you need the trait). I think I'll move the debug builder into just the `test` function in `StarkFriEngine` but move it out of `prove`. Out of scope of this PR.

Closes INT-2914

Added a little more optimization with pre-scanning to avoid checks every row and also don't pack next row unless necessary.


crates/stark-backend/src/air_builders/symbolic/symbolic_expression.rs
```diff
@@ -341,18 +341,64 @@ impl<F: Field> Product<F> for SymbolicExpression<F> {
...
+                    right_idx,
+                    ..
+                } => exprs[left_idx].clone() * exprs[right_idx].clone(),
+                SymbolicExpressionNode::IsFirstRow => self.eval_is_first_row(),
+                SymbolicExpressionNode::IsLastRow => self.eval_is_last_row(),
+                SymbolicExpressionNode::IsTransition => self.eval_is_transition(),
+            };
+            exprs.push(expr);
+        }
+        exprs
```
zlangley, line 402: nit: you can make this a `map`

crates/stark-backend/src/prover/quotient/single.rs
```diff
@@ -58,11 +52,14 @@ where
...
-    let mut alpha_powers = alpha
+    let alpha_powers = alpha
         .powers()
-        .take(symbolic_constraints.constraints.len())
+        .take(constraints.constraint_idx.len())
+        .collect_vec()
+        .into_iter()
+        .rev()
+        .map(PackedChallenge::<SC>::from_f)
         .collect_vec();
```
zlangley, line 62: It would be neat if there were a "reverse collect". oh well.

crates/stark-backend/src/prover/quotient/evaluator.rs
```diff
@@ -0,0 +1,167 @@
...
+        constraints: &SymbolicExpressionDag<Val<SC>>,
+        alpha_powers: &[PackedChallenge<SC>],
+    ) -> PackedChallenge<SC> {
+        let evaluated_nodes = self.eval_nodes(&constraints.nodes);
+        let mut accumulator = PackedChallenge::<SC>::ZERO;
+        for (&alpha_pow, &node_idx) in alpha_powers.iter().zip(&constraints.constraint_idx) {
+            match evaluated_nodes[node_idx] {
+                PackedExpr::Val(x) => {
+                    accumulator += alpha_pow * x;
+                }
```
zlangley, line 161: nit that I thought `cargo fmt` would catch
```suggestion
                PackedExpr::Val(x) => accumulator += alpha_pow * x,
```

zlangley (approved):
Nice

crates/stark-backend/src/prover/quotient/evaluator.rs
jonathanpwang: interesting, indeed it doesn't catch but also does accept

crates/stark-backend/src/air_builders/symbolic/symbolic_expression.rs
jonathanpwang: I tried, but actually it can't because of the `exprs[left_idx]`

zlangley: good point!

crates/stark-backend/src/prover/quotient/evaluator.rs
```diff
@@ -24,10 +24,16 @@ impl<T> ViewPair<T> {
 
     pub fn get(&self, row_offset: usize, column_idx: usize) -> &T {
         match row_offset {
-            0 => &self.local[column_idx],
+            // SAFETY: all column indices have been checked to be in range already
+            0 => unsafe { self.local.get_unchecked(column_idx) },
```
zlangley, line 28: Hmm, `ViewPair` is a public type and `get` is public API. How do we know all this is unsafe in general? We probably should make the method unsafe and then callers can use this function if they have otherwise made the requisite checks. (Or am I missing something?)

jonathanpwang: I changed to pub(crate) afterwards for this reason

nyunyunyunyu:
This is simpler than I thought. LGTM!

crates/stark-backend/src/prover/quotient/evaluator.rs
zlangley: I think it's still easy to mess up and use `get` within this crate in an unsafe way. Isn't it better to mark the function as unsafe and then force the caller to think about it?

PR #20, opened 2025-01-07
[feat] Add `Stateful` trait

nyunyunyunyu:


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "0.2.0-alpha"
+version = "0.2.1-alpha"
```
jonathanpwang, line 2: we haven't released 0.2.0 yet, so ok to not bump for a while

crates/stark-backend/src/chip.rs
```diff
@@ -37,6 +37,12 @@ pub trait ChipUsageGetter {
     }
 }
 
+/// A chip whose state could be persisted.
+pub trait PersistStateChip<S> {
```
jonathanpwang, line 41: I thought you wanted to also implement the trait on some non-chips like `VmChipComplex`? so `Chip` doesn't need to be in the trait name

jonathanpwang (approved):
trait seems fine, let's discuss naming a little

jonathanpwang:
@nyunyunyunyu merge?

jonathanpwang:
hmm @nyunyunyunyu I wonder if we want `load_state` to take in a deserializer so we don't require global changes to switch bincode to bitcode etc?

Cargo.toml
jonathanpwang, line 2:
```suggestion
version = "0.2.0-alpha"
```

PR #21, opened 2025-01-09
chore: move debug out of prover

jonathanpwang:
So `prove` no longer requires `AnyRap`.

I did not remove `AnyRap` from `AirProofInput` because many testing functions use it. I plan to split `AirProofInput` into `SingleKeygenCtx` and `SingleProvingCtx` for the with `AnyRap` and without `AnyRap` variants.

**Note:** this pr is into a `feat/bal` branch, not `main`.

### DIFF
--- a/crates/stark-backend/src/air_builders/debug/check_constraints.rs
+++ b/crates/stark-backend/src/air_builders/debug/check_constraints.rs
@@ -21,11 +21,7 @@ pub fn check_constraints<R, SC>(
     rap_name: &str,
     preprocessed: &Option<RowMajorMatrixView<Val<SC>>>,
     partitioned_main: &[RowMajorMatrixView<Val<SC>>],
-    after_challenge: &[RowMajorMatrixView<SC::Challenge>],
-    challenges: &[Vec<SC::Challenge>],
     public_values: &[Val<SC>],
-    exposed_values_after_challenge: &[Vec<SC::Challenge>],
-    rap_phase_seq_kind: RapPhaseSeqKind,
 ) where
     R: for<'a> Rap<DebugConstraintBuilder<'a, SC>>
         + BaseAir<Val<SC>>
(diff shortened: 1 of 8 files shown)


nyunyunyunyu (approved):

crates/stark-backend/src/air_builders/debug/check_constraints.rs
```diff
@@ -87,14 +68,14 @@ pub fn check_constraints<R, SC>(
                 RowMajorMatrixView::new_row(preprocessed_next.as_slice()),
             ),
             partitioned_main,
-            after_challenge,
-            challenges,
+            after_challenge: vec![], // unreachable
+            challenges: &[],         // unreachable
```
zlangley, line 72: Should we just remove these fields from `DebugConstraintBuilder`?

crates/stark-backend/tests/interaction/mod.rs
```diff
@@ -118,9 +116,6 @@ fn test_interaction_stark_multi_rows_neg() {
         2,
     );
     let receiver_air = DummyInteractionAir::new(1, false, 0);
-    USE_DEBUG_BUILDER.with(|debug| {
-        *debug.lock().unwrap() = false;
-    });
     let res = verify_interactions(
```
zlangley, line 119: No disable here?

crates/stark-backend/tests/interaction/mod.rs
```diff
@@ -216,9 +211,6 @@ fn test_interaction_stark_multi_senders_neg() {
         2,
     );
     let receiver_air = DummyInteractionAir::new(1, false, 0);
-    USE_DEBUG_BUILDER.with(|debug| {
-        *debug.lock().unwrap() = false;
-    });
```
zlangley, line 221: or here?

crates/stark-backend/src/air_builders/debug/check_constraints.rs
jonathanpwang: The folder still needs to implement some traits for `Rap` stuff so it was easier to say its empty here. Also in principle it's general enough we may want it for debugging non-interaction RAPs in the future

crates/stark-backend/tests/interaction/mod.rs
jonathanpwang: `verify_interactions` doesn't use `test`, only `prover` so now that doesn't include the debug builder

crates/stark-backend/tests/interaction/mod.rs
jonathanpwang: ditto

PR #24, opened 2025-01-10
feat: hardware coordinator

jonathanpwang:
- `src/prover/types.rs` has the core types needed by the HAL traits as well as types needed to interface with proving key (which is no hardware specific)
- `src/prover/hal.rs` has traits for hardware data management and device kernels
- `src/prover/coordinator.rs` has the old prove implementation, adapted to use traits so it is just the host cpu coordinator for hardware device implementations of the hard stuff.

Suggested review order:
- `src/prover/mod.rs`
- `src/prover/hal.rs`
- `src/prover/coordinator.rs`
- everything else

Naming conventions (suggested by gpt):
- Buffer: contiguous slice of memory
- Ref: pointer to data only
- View: pointer together with some additional metadata. This should always be cloneable and lightweight, but might have small vecs.

I think the new traits also helped to clean up some of the old code (and hopefully reduce amount of code?).

I arranged the prover to divided as follows (perhaps the verifier should also be structured as such):
- everything before any challenges are needed
- all handling of RAP phases involving challenges: this includes permutation trace generation, commit, and sumcheck. In the future this can include all layered circuits and multi-round sumchecks.
- quotient polynomial calculation and commit. this is only needed in protocols that require quotient polynomial, it can be skipped in full multilinear pcs
- polynomial opening proofs

------
Let's see how well copilot does:
This pull request includes various updates and improvements to the `stark-backend` crate, focusing on dependency updates, type changes, and code simplifications. The most important changes include updating dependencies in `Cargo.toml`, modifying type definitions for interactions, and refactoring the symbolic constraints and DAG structures.

Dependency Updates:
* Updated `derive-new` from version `0.6.0` to `0.7.0` in `Cargo.toml`.
* Added workspace attributes for `derive-new` and modified existing dependencies to use workspace attributes in `crates/stark-backend/Cargo.toml`.

Type and Struct Changes:
* Changed the type of `interactions` from a slice of slices to a slice of vectors in `check_constraints.rs`.
* Updated `SymbolicExpressionNode` and `SymbolicExpressionDag` to include `serialize` and `deserialize` bounds in `dag.rs`. [[1]](diffhunk://#diff-b22f71566fea49bd7c93824105304fe16c62fb9fbcf03aada51c1065b95866beL19-R19) [[2]](diffhunk://#diff-b22f71566fea49bd7c93824105304fe16c62fb9fbcf03aada51c1065b95866beL49-R50)
* Added a new conversion implementation for `SymbolicConstraintsDag` to `SymbolicConstraints` using references in `dag.rs`.

Code Refactoring:
* Removed the `async_trait` and `CircuitProver`, `AsyncCircuitProver`, and `CircuitVerifier` traits from `circuit_api.rs`.
* Removed the `MatrixCommitmentPointers` and related structs from `commit.rs`.
* Refactored the `StarkEngine` trait to use `CpuDevice` for the prover in `engine.rs`.

Simplifications and Improvements:
* Changed `PairTraceView` to use `PairView` with `Arc<RowMajorMatrix>` in `interaction/mod.rs`. [[1]](diffhunk://#diff-c7ea105a2b5b0ea80e0d0d06ae5c47ee443e6857dbe8171b00911e213e3399ccL11-R11) [[2]](diffhunk://#diff-c7ea105a2b5b0ea80e0d0d06ae5c47ee443e6857dbe8171b00911e213e3399ccR188-R189)
* Added `repr(u8)` attribute to `RapPhaseSeqKind` enum in `interaction/mod.rs`.


jonathanpwang:
@zlangley note the prover abstractions described in the PR description. I feel like it may be helpful for how to structure all interaction handling, but I can also address it when I resolve merge conflicts later.

crates/stark-backend/src/circuit_api.rs
```diff
@@ -1,23 +0,0 @@
...
-
-use crate::{
-    config::StarkGenericConfig,
-    prover::types::{Proof, ProofInput},
-    verifier::VerificationError,
-};
-
-/// Async prover for a specific circuit using a specific Stark config.
-#[async_trait]
-pub trait AsyncCircuitProver<SC: StarkGenericConfig> {
```
jonathanpwang, line 11: @nyunyunyunyu I deleted the file because I'm not sure `ProofInput` will be a canonical structure always used, and also this trait implicits that the trace matrix is a solid boundary for prover input, whereas as we've discussed the format of trace may differ depending on the prover/hardware.

I was thinking we can move this into the sdk for now?

nyunyunyunyu:
Can we open a PR in openvm and see if any benchmark downgrade?

jonathanpwang:
@nyunyunyunyu I fixed all tests and added a data transport adapter to `hal.rs` (which is trivial for Cpu).
The conversions from `ProofInput` to `ProvingContext` in `Engine` overall is not very pretty right now. I feel like it should be revisited, but maybe in a separate PR.

crates/stark-backend/src/prover/hal.rs
```diff
@@ -0,0 +1,180 @@
...
+    fn transport_pk_to_device<'a>(
+        &'a self,
+        mpk: &'a MultiStarkProvingKey<SC>,
+        air_ids: Vec<usize>,
+    ) -> MultiStarkProvingKeyView<'a, PB, &'a RapPhaseSeqProvingKey<SC>>
+    where
+        SC: 'a;
+
+    fn transport_matrix_to_device(&self, matrix: &Arc<RowMajorMatrix<Val<SC>>>) -> PB::MatrixView;
+
```
kunxian-xia, line 178: We may also want to have a function to transfer matrix from device to host in the case that we want to have a hybrid prover.

kunxian-xia: But we need to be careful as the side effect of this function implies that the gpu memory occupied by the matrix should be deallocated. This means that we need to deactivate all other `PB::MatrixView`s.

crates/stark-backend/src/prover/hal.rs
```diff
@@ -0,0 +1,180 @@
...
+    type Challenger: CanObserve<Self::Val> + CanObserve<Self::Commitment>;
+
+    // ==== Device Types ====
+    /// Single matrix buffer on device together with dimension metadata.
+    type MatrixView: MatrixView + Clone + Send + Sync;
+    /// Reference for the preimage of a PCS commitment on device, together with any metadata
+    /// necessary for computing opening proofs.
+    ///
+    /// For example, multiple buffers for LDE matrices, their trace domain sizes, and pointer to mixed merkle tree.
+    type PcsDataView: Clone + Send + Sync;
```
kunxian-xia, line 49: I think it's better to define `PcsData` rather than `PcsDataView`. With `PcsData` the prover itself owns the data and can create immutable or mutable references. And the prover can even drop it on gpu and transfer it to host.

crates/stark-backend/src/prover/hal.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    /// The `trace_views` are the views of the respective trace matrices, evaluated on the trace domain.
+    /// Currently this function does not provide a view of any already committed data associated
+    /// with the trace views, although that data is available.
+    fn partially_prove<'a>(
+        &self,
+        challenger: &mut PB::Challenger,
+        pk_views: &[StarkProvingKeyView<'a, PB, Self::RapPartialProvingKeyView<'a>>],
+        trace_views: Vec<PairView<PB::MatrixView, PB::Val>>,
+    ) -> (Self::RapPartialProof, ProverViewAfterRapPhases<PB>);
```
kunxian-xia, line 91: I feel it's better to return the after challenge trace matrix to the prover rather than an reference like `PB::MatrixView`.

crates/stark-backend/src/prover/hal.rs
jonathanpwang: From the host perspective, it should never own `PcsData` though? It will only ever get a reference? I feel like `PcsData` should never leave the device?

crates/stark-backend/src/prover/hal.rs
jonathanpwang: hmm why? I basically never want any large data object (matrix, pcs data, etc) to ever be used on host. the after challenge trace matrix is only used for quotient poly, and we plan for quotient poly to be done entirely on device

crates/stark-backend/src/prover/hal.rs
```diff
@@ -0,0 +1,180 @@
...
+        after_phase: Vec<PB::PcsDataView>,
+        // Quotient poly commitment prover data
+        quotient_data: PB::PcsDataView,
+        // Quotient degree for each RAP committed in quotient_data, in order
+        quotient_degrees: &[u8],
+    ) -> PB::OpeningProof;
+}
+
+/// Trait to manage data transport of prover types from host to device.
+pub trait DeviceDataAdapter<SC, PB>
```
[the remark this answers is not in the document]
jonathanpwang: kk

crates/stark-backend/src/prover/hal.rs
jonathanpwang: @kunxian-xia regarding non-view `Matrix` and `PcsData`, do you want a type for device pointer but where the pointer is fully owned and there's no risk that deallocating would affect other views of it, or do you really want it to backed by memory on the host? I could see a case for the former, but I didn't want the latter because it doesn't seem like a pattern to promote.

jonathanpwang:
Merging into the other feature branch, still open for comments

crates/stark-backend/src/prover/hal.rs
kunxian-xia: Yes, the `PcsData` for GPU backend is almost identical to `DevicePointer`. The rust borrow checker can decide when is immutable or mutable borrow allowed.

kunxian-xia: > we plan for quotient poly to be done entirely on device

It's possible the gpu memory is not enough to hold all matrices and `PcsData` for all chips.

crates/stark-backend/src/prover/hal.rs
kunxian-xia: The `PcsData` for gpu can be `DevicePointer` (the host owns the device object indirectly) and `PcsDataView` can just be `&DevicePointer`. Then you don't need to collect all the `PcsDataView`s and then invoke the `fn deallocate_pcs_datas(pdatas: Vec<PcsDataView>)` when you want to transfer the device's `PcsData` to host.

crates/stark-backend/src/prover/hal.rs
kunxian-xia: This function should work, but you have to collect all `MatrixView` to one `Matrix`, and if one of them is missing, then it might cause GPU OOM issue.

kunxian-xia: If you define associated types like `Matrix` or `PcsData` and when it's out of scope (e.g. call the function `fn transport_matrix_to_host(&self, matrix: PB::Matrix) -> RowMajorMatrix` then the rust compiler can make sure there is no dangling pointers / references.

crates/stark-backend/src/prover/hal.rs
jonathanpwang: OK I see. Let me address it in a follow up PR. I think `PcsData` I basically don't need to have multiple references to. Matrix is different story.

BTW `PcsDataView` is a little more than just traditional `PcsProverData<SC>`: I want it to include ordered list of trace heights within that commit, because you need it for the opening proof.

crates/stark-backend/src/prover/mod.rs
```diff
@@ -1,419 +1,41 @@
...
+    where
+        Self: 'a;
+    type ProvingContext;
+    type Proof;
+
+    /// The prover should own the challenger, whose state mutates during proving.
+    fn prove<'a>(
+        &'a mut self,
+        pk: Self::ProvingKeyView<'a>,
+        ctx: Self::ProvingContext,
```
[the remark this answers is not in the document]
jonathanpwang: why can't we always assume `ProvingContext` contains input?

PR #25, opened 2025-01-10
[fix] calling run_with_metric_collection multiple times fails

Avaneesh-axiom:
Currently, calling the `run_with_metric_collection` multiple times in the same thread, or even across different threads fails. This is because it sets global state by calling `tracing::subscriber::set_global_default` and `metrics::set_global_recorder`. These functions error when called more than once.

The fix is to replace them with `tracing::subscriber::set_default` and `metrics::with_local_recorder` which use local and temporary state instead.

This change allows `run_with_metric_collection` to be used in tests, since we want to be able to run multiple tests.


jonathanpwang (approved):
LGTM after fixing lint

PR #38, opened 2025-02-13
prover: some perf improvement by limiting iterators

tcoratger:


crates/stark-backend/src/prover/coordinator.rs
```diff
@@ -179,31 +180,22 @@ where
...
-                        values.pop();
-                    } else {
-                        break;
-                    }
-                }
-                values
-                    .into_iter()
-                    .map(|v| v.unwrap_or_default())
-                    .collect_vec()
+                    .rev()
```
jonathanpwang, line 194: this seems incorrect? you would need to reverse it back to preserve order

crates/stark-backend/src/prover/coordinator.rs
```diff
@@ -80,24 +80,25 @@ where
         let start = std::time::Instant::now();
         assert!(mpk.validate(&ctx), "Invalid proof input");
 
-        let (air_ids, air_ctxs): (Vec<_>, Vec<_>) = ctx.into_iter().unzip();
-        let num_air = air_ids.len();
-        #[allow(clippy::type_complexity)]
+        let num_air = ctx.per_air.len();
         let (cached_commits_per_air, cached_views_per_air, common_main_per_air, pvs_per_air): (
```
jonathanpwang, line 84: this did not seem to change behavior, and made the code slightly harder to read, so I would prefer not to do it. The types were provided here only because it was sometimes hard for rust-analyzer to decipher them

tcoratger: fixed

crates/stark-backend/src/prover/coordinator.rs
tcoratger: fixed

crates/stark-backend/src/prover/coordinator.rs
```diff
@@ -179,31 +184,26 @@ where
...
         // - transpose per_phase, per_air -> per_air, per_phase
-        let exposed_values_per_air = (0..num_air)
+        let exposed_values_per_air: Vec<Vec<_>> = (0..num_air)
             .map(|i| {
-                let mut values = prover_data_after
+                let mut values: Vec<_> = prover_data_after
                     .rap_views_per_phase
                     .iter()
-                    .map(|per_air| {
+                    .filter_map(|per_air| {
```
jonathanpwang, line 192: this behavior is still a little different than before: if we have `Some, None, Some, None` we want `Some, Default, Some` at the very end. You're giving `Some, Some`. (This case almost never happens, but I'd like to maintain correctness.)

tcoratger: Just reverted, I need to think a bit more to optimize this, the initial goal of my PR was to avoid the double iterator over `ctx.into_iter()` so let us stick to this :)

jonathanpwang (approved):
LGTM

PR #39, opened 2025-02-18
add clippy rules

tcoratger:


jonathanpwang:
Closing for now as it touches too many files. We can revisit after audit is done.

PR #41, opened 2025-02-19
single pass over self.partitioned_airs for constraint deg and symb constraints

tcoratger:


jonathanpwang:
I think this is valid, but it doesn't seem to improve performance much, so let's leave it. Keygen is also security critical so I'd prefer to minimize changes.

PR #43, opened 2025-02-20
Feat: require `dag.constraint_idx` is sorted

kunxian-xia:
## Description
This pr aims to have `dag.constraint_idx` to be sorted.

## Rationale

For complex circuit like keccak, the number of nodes in the dag is very large (around 50k). However, the locality pattern still applies in this case which means each node in the dag only depends on nodes with indices that are close to its index. 

This means we can use a **small** vector for caching the intermediates (we call node like `main / preprocessed / permutation` leaf node, while we call node like `add / mul / sub / neg` as intermediate node). In the special keccak case, we can reduce it from 50k to 200.

If the `constraint_idx` is sorted, then we can free the node's memory (can be reused) after we finished accumulating them into `accumulator`.


jonathanpwang (approved):
LGTM, @nyunyunyunyu can you also review?

I think we expect these idx to be sorted, but apparently for sha2 they aren't because some earlier constraints actually contain the later ones within already.

This affects the vkey, but it is a one line change that seems important, so we will get it audited.

PR #54, opened 2025-03-14
fix: protect against weak Fiat-Shamir

jonathanpwang:
**Finding Link:** https://cantina.xyz/code/c486d600-bed0-4fc6-aed1-de759fd29fa2/findings/152

### Description of Fix

To protect against weak Fiat-Shamir, the prover and verifier should both start off by having the challenger observe something that commits to the full computational statement, i.e., the constraints of the circuit. These are contained in the verifying key. Our fix is to require the honest circuit creator to generate a hash commitment to the constraints and other configuration data at keygen time. This hash commitment is observed at the start of the Fiat-Shamir transcript. Note that the verifier _does not_ need to verify that this is indeed the hash of the verifying key: this can be said to be a property of an honest vkey.

Due to the existing types and what's available in the code, the hash commitment is done by first serializing vkey to bytes via `bitcode`, converting each byte to a field element, and then considering as a 1 column matrix and using the `pcs` to commit to it. I had wanted to use `bincode v2.0.0` but there is a MSRV of 1.85. Since this hash doesn't need to be checked by the verifier, the use of `bitcode` seems acceptable. The use of `pcs` has the nice(?) side benefit that it implicitly commits to the blowup factor.

Towards INT-3564

### DIFF
--- a/crates/stark-backend/src/keygen/mod.rs
+++ b/crates/stark-backend/src/keygen/mod.rs
@@ -2,9 +2,10 @@ use std::{collections::HashMap, iter::zip, sync::Arc};
 
 use itertools::Itertools;
 use p3_commit::Pcs;
-use p3_field::{Field, FieldExtensionAlgebra};
-use p3_matrix::Matrix;
+use p3_field::{Field, FieldAlgebra, FieldExtensionAlgebra};
+use p3_matrix::{dense::RowMajorMatrix, Matrix};
 use tracing::instrument;
+use types::MultiStarkVerifyingKey0;
 
 use crate::{
     air_builders::symbolic::{get_symbolic_builder, SymbolicRapBuilder},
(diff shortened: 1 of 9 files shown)


crates/stark-backend/src/keygen/mod.rs
```diff
@@ -195,11 +196,33 @@ impl<'a, SC: StarkGenericConfig> MultiStarkKeygenBuilder<'a, SC> {
...
 
+        let pre_vk: MultiStarkVerifyingKey0<SC> = MultiStarkVerifyingKey0 {
+            per_air: pk_per_air.iter().map(|pk| pk.vk.clone()).collect(),
+            trace_height_constraints: trace_height_constraints.clone(),
+            log_up_pow_bits: log_up_security_params.log_up_pow_bits,
+        };
+        // To protect against weak Fiat-Shamir, we hash the "pre"-verifying key and include it in the
+        // final verifying key. This just needs to commit to the verifying key and does not need to be
+        // verified by the verifier, so we just use bincode to serialize it.
+        let vk_bytes = bitcode::serialize(&pre_vk).unwrap();
```
zlangley, line 207: Noting that if we change bitcode versions this may change.

crates/stark-backend/src/keygen/mod.rs
```diff
@@ -195,11 +196,33 @@ impl<'a, SC: StarkGenericConfig> MultiStarkKeygenBuilder<'a, SC> {
...
+        tracing::info!("pre-vkey: {} bytes", vk_bytes.len());
+        // Purely to get type compatibility and convenience, we hash using pcs.commit as a single row
+        let vk_as_row = RowMajorMatrix::new_row(
+            vk_bytes
+                .into_iter()
+                .map(Val::<SC>::from_canonical_u8)
```
zlangley, line 213: Observation: The commitment can be made a bit faster if we pack three bytes into the field.

crates/stark-backend/src/keygen/types.rs
```diff
@@ -87,6 +87,20 @@ pub struct MultiStarkVerifyingKey<SC: StarkGenericConfig> {
...
+    /// initialize by observing this hash.
+    pub pre_hash: Com<SC>,
+}
+
+#[derive(Serialize, Deserialize)]
+#[serde(bound(
+    serialize = "Com<SC>: Serialize",
+    deserialize = "Com<SC>: Deserialize<'de>"
+))]
+pub(crate) struct MultiStarkVerifyingKey0<SC: StarkGenericConfig> {
```
zlangley, line 100: This means every time we want to add a field to the verifying key we need to remember to add it here too, right? Can we add a const assert that `size_of(MultiStarkVerifyingKey0) == size_of(MultiStarkVerifyingKey) - size_of(Com)` or something?

zlangley (approved):
It would be nice if only adding a field to `MultiStarkVerifyingKey` and not `MultiStarkVerifyingKey0` resulted in a compilation issue. Added a suggestion inline.

crates/stark-backend/src/keygen/mod.rs
jonathanpwang: yea this is why I wanted bincode v2.0.0 (more stable), I'm also open to just fixing it to bincode v1.3.3 for more stability?

crates/stark-backend/src/keygen/mod.rs
jonathanpwang: yea... a bit hacky so I skipped it

crates/stark-backend/src/keygen/types.rs
jonathanpwang: interesting

jonathanpwang: the generic `SC` was a blocker

crates/stark-backend/src/keygen/mod.rs
jonathanpwang: switched to `bincode`

crates/stark-backend/src/keygen/types.rs
jonathanpwang: switched to `inner` pattern

crates/stark-backend/src/keygen/mod.rs
jonathanpwang: I think it's ok since this is not really externally used and does not need to be reconstructable after keygen

PR #57, opened 2025-03-23
fix: verifier should check proof shape

jonathanpwang:
A good guiding principle was that `challenge.observe_slice` does not observe the length of a slice, so for every use of observe slice where we don't observe length separately, it should mean that the verifier must validate that the length matches an expected length in the vkey.

### DIFF
--- a/crates/stark-backend/src/verifier/mod.rs
+++ b/crates/stark-backend/src/verifier/mod.rs
@@ -41,7 +41,7 @@ impl<'c, SC: StarkGenericConfig> MultiTraceStarkVerifier<'c, SC> {
         mvk: &MultiStarkVerifyingKey<SC>,
         proof: &Proof<SC>,
     ) -> Result<(), VerificationError> {
-        // Note: construction of view panics if any air_id exceeds num_airs
+        // Note: construction of view panics if any air_id exceeds number of AIRs in provided `MultiStarkVerifyingKey`
         let mvk = mvk.view(&proof.get_air_ids());
         self.verify_raps(challenger, &mvk, proof)?;
         Ok(())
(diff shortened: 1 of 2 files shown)


crates/stark-backend/src/verifier/mod.rs
```diff
@@ -39,6 +41,7 @@ impl<'c, SC: StarkGenericConfig> MultiTraceStarkVerifier<'c, SC> {
         mvk: &MultiStarkVerifyingKey<SC>,
         proof: &Proof<SC>,
     ) -> Result<(), VerificationError> {
+        // Note: construction of view panics if any air_id exceeds num_airs
```
jonathanpwang, line 44:
```suggestion
        // Note: construction of view panics if any air_id exceeds number of AIRs in `mvk`
```

crates/stark-backend/src/verifier/mod.rs
```diff
@@ -74,16 +78,27 @@ impl<'c, SC: StarkGenericConfig> MultiTraceStarkVerifier<'c, SC> {
...
             }
         }
-        // Check that all `air_id`s are different
+        // (T01a): Check that all `air_id`s are different and contained in `MultiStarkVerifyingKey`
         {
             let mut air_ids = air_ids;
             air_ids.sort();
             for ids in air_ids.windows(2) {
-                assert!(ids[0] < ids[1], "all `air_id`s must be different");
+                if ids[0] >= ids[1] {
```
nyunyunyunyu, line 86: If `ids[0] > ids[1]`, technically it's not `DuplicateAirs`

nyunyunyunyu (commented):
LGTM. Rust verifier needs less checks than the DSL verifier because Rust has out-of-bound check by default.

crates/stark-backend/src/verifier/mod.rs
jonathanpwang: true, but we also just sorted it, so actually we could have done `if ids[0] == ids[1]`

PR #58, opened 2025-03-24
chore: Make SymbolicExpressionDag fields public

gaxiom:


crates/stark-backend/src/air_builders/symbolic/dag.rs
```diff
@@ -65,6 +65,10 @@ impl<F> SymbolicExpressionDag<F> {
         }
         rotation
     }
+
```
gaxiom, line 68: This one isn't necessary after making constraint_idx public. I can delete it
@jonathanpwang

PR #71, opened 2025-04-30
feat!: drop main trace earlier for lower peak memory

jonathanpwang:
Changed the traits slightly so that traces are passed without borrowing. The main reason is the realization that preprocessed and main traces are only used to generate the permutation trace, so they can be dropped immediately after this step (in `partially_prove`). Afterwards, only LDEs need to be used.

There is some improvement in how the handling of preprocessed / cached trace is done: sometimes you may want to not drop these (so they really stay in memory) and sometimes you may want to drop it to lower peak memory. Maybe some `Shared**` type may make this clearer, but for now I just opted to use owned versions of everything because in practice we are using smart pointers (`Arc`) around everything, so one can get around sharing with some cloning if necessary.

Note that right now the preprocessed trace is never dropped, because the proving key holds onto a copy until `Opening` is called. We could get around this with some extra `Option` usage, but since the size of preprocessed trace is currently small I'll leave it for later.

https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/14749714037/job/41404170207

### DIFF
--- a/crates/stark-backend/src/engine.rs
+++ b/crates/stark-backend/src/engine.rs
@@ -1,4 +1,4 @@
-use std::{iter::zip, sync::Arc};
+use std::{iter::zip, marker::PhantomData, sync::Arc};
 
 use itertools::{zip_eq, Itertools};
 use p3_matrix::Matrix;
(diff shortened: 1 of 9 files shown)


gaxiom (approved):
LGTM

It will help to control PcsData at any moment which is nice.

PR #73, opened 2025-05-01
feat: reserve capacity for perm & quotient LDE; avoid copy for quotient split

jonathanpwang:
Follow up of https://github.com/openvm-org/stark-backend/pull/68 where the main idea was that for perm and quotient PCS commitments, we generate the matrices and then immediate apply cosetLDE which does in-place DFT which resizes the matrix to `trace_height * log_blowup`. To avoid re-allocation, we allocate the right amount of space from the start.

While doing this, I noticed that we were using plonky3's `Domain::split_evals` to split the quotient poly into chunks: this does a full copy of the quotient matrix (basically a transpose) -- this is totally unnecessary since we generate the whole quotient poly using a `par_iter`! Under (what I think is) a minor assumption that we always split quotient polys in this way, I got rid of it so now there's no copy.

https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/14790290465
513s -> 500s total prove time

### DIFF
--- a/crates/stark-backend/src/interaction/fri_log_up.rs
+++ b/crates/stark-backend/src/interaction/fri_log_up.rs
@@ -23,13 +23,17 @@ use crate::{
 
 pub struct FriLogUpPhase<F, Challenge, Challenger> {
     log_up_params: LogUpSecurityParameters,
+    /// When the perm trace is created, the matrix will be allocated with `capacity = trace_length << extra_capacity_bits`.
+    /// This is to avoid resizing for the coset LDE.
+    extra_capacity_bits: usize,
     _marker: PhantomData<(F, Challenge, Challenger)>,
 }
 
 impl<F, Challenge, Challenger> FriLogUpPhase<F, Challenge, Challenger> {
-    pub fn new(log_up_params: LogUpSecurityParameters) -> Self {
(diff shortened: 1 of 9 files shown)


crates/stark-backend/src/interaction/fri_log_up.rs
```diff
@@ -329,7 +337,10 @@ where
         // on the fly. If we introduce a more advanced chunking algorithm, then we will need to
         // cache the chunking information in the proving key.
         let perm_width = interaction_partitions.len() + 1;
-        let mut perm_values = Challenge::zero_vec(height * perm_width);
+        // We allocate extra_capacity_bits now as it will be needed by the coset_lde later in pcs.commit
+        let perm_trace_len = height * perm_width;
+        let mut perm_values = Challenge::zero_vec(perm_trace_len << extra_capacity_bits);
+        perm_values.truncate(perm_trace_len);
```
gaxiom, line 343: Why do you need to truncate it?

gaxiom (approved):
LGTM

crates/stark-backend/src/interaction/fri_log_up.rs
jonathanpwang: truncate sets `vec.len()` without changing `vec.capacity()`

PR #75, opened 2025-05-05
perf: optimize `compute_quotient_values`

jonathanpwang:
This is a PR towards optimizations of the quotient values computation.

I have added a new example for benchmark/profiling purposes as well as a readme in the `quotient/` folder explaining our current approach and some ideas on where to optimize.

- [ ] Add example that uses `p3-uni-stark` to compare against the trait approach

1. `par_chunks` -> `parallelize_chunks` https://github.com/openvm-org/stark-backend/commit/979aba570e2b479706a7b6b8668d06ee6364c318 ([reth bench](https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/14827865541/job/41623727390)) - seems better with low thread count but worse with high
2. allocate new vecs only once per thread instead of each par_iter (builds on 1) 2d9f814162c4847252c9e24bdee7ebe06b1d39d9 ([reth bench](https://github.com/axiom-crypto/openvm-reth-benchmark/blob/gh-pages/benchmarks-dispatch/refs/heads/main/reth-194eefe747af7f8e3e147db2d1d8543d942c645c-d72e259f9153d8a6b647426f5f463368636de110d8fbad9a8c26f94834d67321.md)) - improvement over 1 and with some more use of unsafe got better than `main` (488s) though it's possible some of the tricks could have been done without (1)


jonathanpwang:
@zlangley I am weirdly seeing that https://github.com/openvm-org/stark-backend/blob/a755c4f118d91cd545f2fd21c38c03ebe7acc43d/crates/stark-backend/src/prover/cpu/quotient/single.rs#L88 takes a lot of time, if I'm reading the flamegraph correctly

zlangley:
> @zlangley I am weirdly seeing that
> 
> https://github.com/openvm-org/stark-backend/blob/a755c4f118d91cd545f2fd21c38c03ebe7acc43d/crates/stark-backend/src/prover/cpu/quotient/single.rs#L88
> 
> takes a lot of time, if I'm reading the flamegraph correctly

It might be because there is no parallelism in `selectors_on_coset`. I believe the expensive work in parallel tasks is hard to identify, even in samply.

PR #88, opened 2025-08-15
feat: New gpu crates

gaxiom:


crates/cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,54 @@
+[package]
+name = "openvm-cuda-backend"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+description = "CUDA backend for BabyBear Poseidon2 config"
```
jonathanpwang, line 6:
```suggestion
description = "OpenVM CUDA prover backend for BabyBear Poseidon2 config"
```

crates/cuda-backend/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+p3-merkle-tree = { workspace = true }
+p3-symmetric = { workspace = true }
+p3-fri = { workspace = true }
+p3-dft = { workspace = true }
+
+[build-dependencies]
+cc = { version = "1.0", features = ["parallel"] }
+
+[dev-dependencies]
+clap = { version = "4.4", features = ["derive"] }
```
jonathanpwang, line 44: still better to put in workspace

crates/cuda-backend/examples/keccak256.rs
jonathanpwang, line 1: discussed offline: we will move this to openvm to avoid circular repo dependency

crates/cuda-common/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+tracing.workspace = true
+metrics.workspace = true
+lazy_static.workspace = true
+thiserror.workspace = true
+bytesize = "2.0"
```
jonathanpwang, line 16: workspace

Cargo.toml
```diff
@@ -8,7 +8,9 @@ repository = "https://github.com/openvm-org/"
 license = "MIT OR Apache-2.0"
 
 [workspace]
-members = ["crates/stark-backend", "crates/stark-sdk"]
+members = ["crates/stark-backend", "crates/stark-sdk", "crates/cuda-common", "crates/cuda-backend", "crates/cuda-builder"]
+# Only these build by default
+default-members = ["crates/stark-backend", "crates/stark-sdk"]
```
jonathanpwang, line 13: is it the case that you could actually include all members as default now that there is `cfg(has_cuda)`?

crates/cuda-backend/examples/keccakf.rs
```diff
@@ -1,5 +1,9 @@
+#![cfg_attr(not(has_cuda), allow(unused_imports, dead_code))]
+
 use std::sync::Arc;
 
+#[cfg(has_cuda)]
```
jonathanpwang, line 5: I feel like we should just put `#![cfg(has_cuda)]` at the top, and then in Cargo.toml I think there is a way to specify certain binaries require some cfg

gaxiom: If you think it's better - I can have a feature and still have availability check:
```Rust
if cuda_available() {
        println!("cargo:rustc-cfg=feature=\"cuda\"");
    }
```

jonathanpwang: wow didn't know you could do that -- yea that is ideal

Cargo.toml
jonathanpwang: I think if you are in a cuda machine, right now `cargo check` won't check the cuda crates because they aren't default

jonathanpwang (approved):
LGTM, you can merge after the last changes we were discussing.

This was a beautiful PR

PR #100, opened 2025-08-20
chore: script + workflow to run clang-tidy on CUDA code

stephenh-axiom-xyz:
Resolves INT-4738


crates/cuda-backend/cuda/src/prefix.cu
```diff
@@ -7,7 +7,13 @@ static const uint32_t SHARED_DATA = 256;
 __global__ void prefix_scan_block_ext(FpExt *d_inout, uint64_t length, uint64_t round_stride) {
     FpExt acc_res = FpExt(0);
     FpExt acc_data[ACC_PER_THREAD];
+
+#ifndef __clang_analyzer__
```
gaxiom, line 11: It's not possible to teach clang to allow this?

gaxiom (approved):

PR #101, opened 2025-08-20
chore(cuda): Add files and includes to watch list

gaxiom:


crates/cuda-common/build.rs
```diff
@@ -8,7 +8,6 @@ fn main() {
 
         let include_path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap()).join("include");
         println!("cargo:include={}", include_path.display()); // -> DEP_CUDA_COMMON_INCLUDE
-        println!("cargo:rerun-if-changed={}", include_path.display());
```
jonathanpwang, line 11: why was this deleted?

gaxiom: We should not depend on CARGO_MANIFEST_DIR for sure, but I guess we should depend on local include

PR #103, opened 2025-08-25
fix: use plonky3 `Packed` types for SIMD compatibility

jonathanpwang:
We don't use the type itself, but we need to use the correct type to avoid compile errors.

The `ValMmcs` has packed types which is phantom data and only used by the prover (which we ignore since we use the cuda prover), and the proof types only used the associated `::Value` which is SIMD-independent.

closes INT-4848


shuklaayush (approved):
i guess it's fine since it compiles now and tests pass

PR #106, opened 2025-08-25
chore: Add the signature matcher, fix the one warning it found

Golovanov399:
This resolves INT-4796.

It probably doesn't find as much as it could because cbindgen needs `pub` to see a function. But it found one place where Rust provided a `usize` and cuda read a `uint32_t`, so seems working


gaxiom (approved):
Can't say that I truly understand what it does.
If it works - ok

PR #110, opened 2025-08-27
fix: cuda memory init

gaxiom:


jonathanpwang (approved):
LGTM, but only merge to main after confirming openvm and openvm-reth-benchmark work

PR #114, opened 2025-09-01
feat(cuda): Virtual Memory Manager

gaxiom:


jonathanpwang:
Also requested @Golovanov399 to review since he is the one who first brought up memory fragmentation

.github/workflows/lints.yml
```diff
@@ -80,8 +80,3 @@ jobs:
           cargo clippy --all-targets --all --tests --features "default mimalloc nightly-features parallel prometheus touchemall" -- -D warnings
           cargo clippy --all-targets --all --tests --no-default-features --features "jemalloc jemalloc-prof" -- -D warnings
 
-      - name: Run clang-tidy
```
jonathanpwang, line 83: please put this back; should be fixed now if you rebase main

crates/cuda-builder/src/lib.rs
```diff
@@ -18,12 +18,9 @@ pub struct CudaBuilder {
 impl Default for CudaBuilder {
     fn default() -> Self {
         let mut link_search_paths = Vec::new();
-        if let Ok(ld_path) = env::var("LD_LIBRARY_PATH") {
-            for path in ld_path.split(':') {
-                if !path.is_empty() {
-                    link_search_paths.push(path.to_string());
-                }
-            }
+        link_search_paths.push("/usr/local/cuda/lib64".to_string());
```
jonathanpwang, line 21: can you undo this? not good to use `/usr/local/cuda` now that I realize we may have multiple cuda versions installed and explicitly don't want to go through the symlink

crates/cuda-builder/src/lib.rs
```diff
@@ -18,12 +18,8 @@ pub struct CudaBuilder {
 impl Default for CudaBuilder {
     fn default() -> Self {
         let mut link_search_paths = Vec::new();
-        if let Ok(ld_path) = env::var("LD_LIBRARY_PATH") {
```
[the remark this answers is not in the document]
gaxiom: If you think that `/usr/local/cuda` is dangerous because of multiple versions on device, better to have a controled env var for that

crates/cuda-common/VPMM_SPEC.md
jonathanpwang, line 1: should we move this to `./docs` and link from README.md in this crate?

if you prefer keeping it here, make README.md and add a link

crates/cuda-builder/src/lib.rs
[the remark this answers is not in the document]
jonathanpwang: I think we should have some default for convenience, so that can be `/usr/local/cuda` if that's what you prefer, but then we should document how to override

crates/cuda-common/src/memory_manager/vm_pool.rs
```diff
@@ -0,0 +1,325 @@
...
+            return Err(MemoryError::NotInitialized);
+        }
+        assert!(
+            requested != 0 && requested % self.page_size == 0,
+            "Requested size must be a multiple of the page size"
+        );
+        // Try to find best fit free region
+        let mut best_region = self
+            .free_regions
+            .iter()
```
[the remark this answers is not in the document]
gaxiom: Yes. A tradeoff between:
- always on malloc - try to find smallest region ≥ size
- always on free - try to find left and right neighbour

I decided not to keep same mapping in two structs (which is the best performance) because of annoying overhead of making them always identical

crates/cuda-common/VPMM_SPEC.md
jonathanpwang: yes

PR #119, opened 2025-09-04
fix(cuda): Stop using constant twiddle

gaxiom:
Closes INT-4949

@jonathanpwang let's verify with openvm patch first - then merge

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace.package]
-version = "1.2.1-rc.0"
+version = "1.2.1-rc.1"
 edition = "2021"
 rust-version = "1.83"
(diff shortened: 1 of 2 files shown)


jonathanpwang:
@gaxiom I will wait for you to patch and check in openvm

jonathanpwang:
Running a comparison: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/17504139965

jonathanpwang:
I bumped workspace to version rc.1 -- in the future we can switch to `develop` branch for optimizations, but this is more of a stability fix so I think we should tag it.

jonathanpwang (approved):

PR #124, opened 2025-09-10
chore: use newer ami

luffykai:


jonathanpwang:
CI fails because it uses runner from `main`

jonathanpwang (approved):

PR #130, opened 2025-09-14
chore(cuda): universal babybear impl

gaxiom:
Closes INT-4987


crates/cuda-common/include/fpext.h
```diff
@@ -41,191 +37,109 @@
 /// smallest B which makes the polynomial irreducible.
 struct FpExt {
     /// The elements of FpExt, elems[0] + elems[1]*X + elems[2]*X^2 + elems[3]*x^4
-    Fp elems[4];
-
+    union {
```
jonathanpwang, line 40: why?

gaxiom: Best way to use `bb31_4_t` without changing access to Fp through elems[i]

gaxiom:
> Currently failing:
> 
> ```
> GPU free failed: Cuda(CudaError { code: 700, name: "cudaErrorIllegalAddress", message: "an illegal memory access was encountered" })
> ```

Fixed. It was an issue with usize > uint32_t mapping 
https://github.com/openvm-org/openvm/pull/2145

PR #135, opened 2025-09-25
perf(cuda): Opener in natural order

gaxiom:
Closes INT-4618

Making all `reduced_openings` polynomial & `z_diff_invs` in natural order.
- We still use z_diff_inv for max height only => double bit reverse required for smaller matrices
- Also I've changed some buffers with one value to just values

### DIFF
--- a/crates/cuda-backend/cuda/src/fri.cu
+++ b/crates/cuda-backend/cuda/src/fri.cu
@@ -14,8 +14,8 @@ __forceinline__ __device__ uint32_t bit_rev(uint32_t x, uint32_t n) {
     return __brev(x) >> (__clz(n) + 1);
 }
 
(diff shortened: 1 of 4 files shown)


jonathanpwang (commented):
Comparison benchmark?

jonathanpwang (commented):
Please give PR description explaining your changes

gaxiom:
> Comparison benchmark?

You can compare this: https://github.com/openvm-org/openvm/pull/2149#issuecomment-3336563573 with this: https://github.com/openvm-org/openvm/pull/2149#issuecomment-3336032686
Also I've started reth compare: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/18026057090
that we can compare with just babybear: https://github.com/axiom-crypto/openvm-reth-benchmark/actions/runs/18019116139

PR #304, opened 2026-03-20
feat: update root system params for efficient verifier

jonathanpwang:
Making the root params closer to the old compression params. We need the verifier cost to be as small as possible for the static verifier.

closes INT-6650


jonathanpwang:
@stephenh-axiom-xyz this is ready

stephenh-axiom-xyz (approved):

PR #307, opened 2026-03-24
fix(cuda): bn254 prover performance

gaxiom:
Closes INT-6989
Closes INT-7001

- Critical correctness fix (NUM_F_ELMS 7→8) is the most important change — this was the root cause of GPU/CPU Merkle hash divergence
- Clean removal of the CPU fallback (new_bn254_host) now that GPU hashing is correct

### DIFF
--- a/crates/cuda-backend/src/bn254_sponge.rs
+++ b/crates/cuda-backend/src/bn254_sponge.rs
@@ -320,8 +320,7 @@ impl GpuFiatShamirTranscript<BabyBearBn254Poseidon2Config> for MultiField32Chall
 
 #[cfg(test)]
 mod tests {
-    use openvm_stark_backend::FiatShamirTranscript;
-    use openvm_stark_backend::p3_symmetric::CryptographicHasher;
+    use openvm_stark_backend::{p3_symmetric::CryptographicHasher, FiatShamirTranscript};
     use openvm_stark_sdk::config::baby_bear_bn254_poseidon2::default_transcript;
     use p3_field::PrimeCharacteristicRing;
     use p3_symmetric::MultiField32PaddingFreeSponge;
@@ -598,11 +597,10 @@ mod tests {
     fn test_grind_gpu_matches_cpu_after_large_commit_observes() {
         let mut gpu = MultiField32ChallengerGpu::new();
         let mut cpu = default_transcript();
-        let hash =
-            MultiField32PaddingFreeSponge::<BabyBear, Bn254Scalar, _, 3, 16, 1>::new(
-                default_babybear_bn254_poseidon2(),
-            )
-            .unwrap();
+        let hash = MultiField32PaddingFreeSponge::<BabyBear, Bn254Scalar, _, 3, 16, 1>::new(
+            default_babybear_bn254_poseidon2(),
+        )
+        .unwrap();
 
         for seed in [17u32, 113, 997] {
             let digest = hash.hash_slice(
--- a/crates/cuda-backend/src/bn254_sponge.rs
+++ b/crates/cuda-backend/src/bn254_sponge.rs
@@ -321,8 +321,10 @@ impl GpuFiatShamirTranscript<BabyBearBn254Poseidon2Config> for MultiField32Chall
 #[cfg(test)]
 mod tests {
     use openvm_stark_backend::FiatShamirTranscript;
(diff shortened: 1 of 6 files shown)


stephenh-axiom-xyz (commented):
From the continuation test times in your OpenVM PR vs. one from about a week ago [here](https://github.com/openvm-org/openvm/actions/runs/23312482425/job/67803154611) this is still a regression. See the comment below.
In yours:
```
PASS [  50.541s] ( 2/17) openvm-continuations tests::test_deferral_aggregation::partially_empty_aggregation_tree
```
In the one linked:
```
PASS [  34.131s] ( 2/11) openvm-continuations tests::test_deferral_aggregation::partially_empty_aggregation_tree
```

crates/cuda-backend/src/merkle_tree.rs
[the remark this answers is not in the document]
gaxiom: It is. And I'll fix it rn

crates/cuda-backend/src/merkle_tree.rs
```diff
@@ -228,14 +186,6 @@ impl<D: Copy + Send + Sync + 'static> MerkleTreeGpu<F, D> {
...
         MerkleTreeError,
     > {
-        if backing_matrices.len() > 1 {
-            return Self::batch_open_rows_host(
-                backing_matrices,
-                query_indices,
-                query_stride,
-                rows_per_query,
-            );
-        }
```
zlangley, line 238: Thanks!

crates/cuda-backend/src/merkle_tree.rs
```diff
@@ -397,65 +274,10 @@ impl MerkleTreeGpu<F, Digest> {
 
 // Base field merkle tree — generic batch query (works for any BatchQueryMerkle digest)
 impl<D: BatchQueryMerkle + Send + Sync + 'static> MerkleTreeGpu<F, D> {
-    #[cfg(feature = "baby-bear-bn254-poseidon2")]
-    fn batch_query_merkle_proofs_host(
+    fn batch_query_merkle_proofs_device(
```
zlangley: I think we can remove `_device` suffix since it's on `MerkleTreeGpu`?

crates/cuda-backend/src/merkle_tree.rs
```diff
@@ -532,6 +354,38 @@ impl MerkleProofQueryDigest for Digest {
...
+        MerkleTreeError,
+    >
+    where
+        D: MerkleProofQueryDigest,
+    {
+        D::batch_query_merkle_proofs(trees, query_indices)
+    }
+}
+
+impl MerkleProofQueryDigest for Digest {
```
stephenh-axiom-xyz, line 382: nit: as mentioned in the Claude review, it may be possible to do this for a generic digest

stephenh-axiom-xyz (approved):
LGTM

stephenh-axiom-xyz (commented):
From Slack we need to take a closer look at things

crates/cuda-backend/src/bn254_sponge.rs
```diff
@@ -591,4 +592,49 @@ mod tests {
 
         run_steps(&steps);
     }
+
+    #[test]
+    fn test_grind_gpu_matches_cpu_after_large_commit_observes() {
+        let mut gpu = MultiField32ChallengerGpu::new();
+        let mut cpu = default_transcript();
+        let hash = MultiField32PaddingFreeSponge::<BabyBear, Bn254Scalar, _, 3, 16, 1>::new(
```
zlangley: We don't use this sponge for the transcript, so I think this test is kind of misleading.

zlangley (approved):

crates/cuda-backend/src/bn254_sponge.rs
gaxiom: So, do you think that I should just remove this test? @zlangley

zlangley: I think all the tests in this file are a little weird in that the only way we actually use this config is to hash, but in this file we use it as the sponge for a transcript (duplex interface). At the same time it does establish parity between the GPU and Plonky3 implementations.

zlangley: I don't think we should remove the test. It would be nice if we could move off of the transcript interface. Don't feel that strongly about it though, it just confused me when I was reviewing.

PR #312, opened 2026-03-30
fix: `u16` overflow of `bus_index + 1`

Golovanov399:
This resolves INT-7120


zlangley (approved):

jonathanpwang:
@claude do we need to make a matching fix in the cuda prover, cpu prover, or reference cpu prover?

PR #315, opened 2026-03-31
fix(cuda): stream for defrag leftovers 

gaxiom:
Closes INT-7062

### DIFF
--- a/crates/cuda-common/src/memory_manager/vm_pool.rs
+++ b/crates/cuda-common/src/memory_manager/vm_pool.rs
@@ -850,7 +850,7 @@ mod tests {
     use std::sync::Arc;
 
     use super::{FreeRegionMeta, VirtualMemoryPool, VpmmConfig};
-    use crate::stream::{CudaEvent, CudaStream, current_stream_id};
+    use crate::stream::{current_stream_id, CudaEvent, CudaStream};
 
     #[test]
     fn test_defrag_leftover_preserves_original_metadata() {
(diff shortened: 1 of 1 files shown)


jonathanpwang:
@codex review

crates/cuda-common/src/memory_manager/vm_pool.rs
```diff
@@ -833,3 +844,57 @@ impl std::fmt::Debug for VirtualMemoryPool {
...
+        let mut pool = VirtualMemoryPool::new(config);
+
+        if pool.page_size == usize::MAX {
+            println!("VPMM not supported, skipping test");
+            return;
+        }
+
+        let page_size = pool.page_size;
+        let stream_id = current_stream_id().unwrap();
+        let foreign_stream_id = stream_id + 1000;
```
jonathanpwang: shouldn't foreign_stream_id be determined from foreign_stream?

jonathanpwang (approved):
the fix itself looks good, but I had a question about the test

crates/cuda-common/src/memory_manager/vm_pool.rs
jonathanpwang: ok apparently it's hard to get stream id from the CudaStream -- request to just rename `foreign_stream_id` to `placeholder_stream_id`

gaxiom: We now have only "current_stream_id" - we don't have a method that returns id for any stream

PR #317, opened 2026-04-04
chore(cuda): stream migration refactoring

gaxiom:
Closes INT-6464

## CUDA Explicit Stream Migration

Replace implicit per-thread default stream (PTDS) with explicit `cudaStream_t` management across the entire CUDA stack. Removes the `--default-stream=per-thread` compiler flag, introduces `CudaStream`, `StreamGuard`, and `DeviceContext` abstractions, and threads an explicit stream through every kernel launch, memory operation, and copy path.

**87 files changed, +3750 / -1758**

See [cuda-stream-migration-design.md](https://github.com/openvm-org/v2-proof-system/blob/test/stream-migration/cuda-stream-migration-design.md) for full design rationale.

---

### `ProverDevice` trait — `DeviceCtx` associated type

`ProverDevice` gains an associated type and accessor. GPU devices return `DeviceContext` (CUDA stream + device ID), CPU/Reference devices return `()`:

```rust
pub trait ProverDevice<PB, TS>: ... {
    type DeviceCtx: Clone + Send + Sync;
    fn device_ctx(&self) -> &Self::DeviceCtx;
}
```

A convenience alias `EngineDeviceCtx<E>` is added to `engine.rs` to avoid spelling out the full associated type path.

---

### `stream.rs` — New types

- **`CudaStream`** — RAII wrapper around `cudaStream_t`, created with `cudaStreamNonBlocking`. Non-blocking `Drop` (calls `cudaStreamDestroy` only, no sync). Per-stream `Mutex<CudaEvent>` for D2H sync via `to_host_sync()`.
- **`StreamGuard`** — newtype around `Arc<CudaStream>` for ref-counted stream ownership. `is_same_stream()` via `Arc::ptr_eq`. `Deref<Target=CudaStream>`.
- **`DeviceContext`** — thin `(device_id, StreamGuard)` bundle. Created via `DeviceContext::for_device(id)` which binds the CUDA device before stream creation.

### `memory_manager/` — Stream-aware allocation records

- `AllocRecord { size, stream: StreamGuard }` — tracks which stream each small allocation belongs to
- `VpmmRecord { size, stream: StreamGuard }` — same for VPMM large allocations
- Two-stage `d_free`: bookkeeping + `cudaFreeAsync` under lock, `StreamGuard` dropped outside lock
- VPMM same-stream matching via `Arc::ptr_eq`; cross-stream reuse gated by event completion

### `copy.rs` / `d_buffer.rs` — `_on(&DeviceContext)` API

All legacy streamless APIs removed. Only `_on(&DeviceContext)` variants remain: `to_device_on`, `to_host_on`, `fill_zero_on`, `with_capacity_on`, `d_malloc_on`, `cuda_memcpy_on`. `DeviceBuffer` gains `#[cfg(feature = "debug-cuda-stream")] alloc_stream` field for wrong-stream detection.

### Removed

`cudaStreamPerThread`, `CudaStreamId`, `current_stream_id()`, `current_stream_sync()`, `default_stream_wait()`, `record_on_this()`, `CudaThreadCleanup`, `gpu_metrics_span()` (PTDS variant), `CudaStream::new()` (blocking), global `COPY_EVENT` mutex, `--default-stream=per-thread` flag.

---

### Device model (`device.rs`, `engine.rs`)

- `GpuDevice` renamed to `GpuDeviceConfig` (pure config)
- New `GpuDevice { config: GpuDeviceConfig, device_ctx: DeviceContext }` — stream-owning device handle
- `GpuEngine::new()` calls `GpuDevice::new().expect(...)` to preserve infallible `StarkEngine::new()`
- `ProverDevice::DeviceCtx = DeviceContext`, `device_ctx()` returns `&self.device_ctx`

### CUDA launchers (25 `.cu` files)

Every `extern "C"` launcher gains `cudaStream_t stream` as final parameter. All kernel launches use `<<<grid, block, 0, stream>>>`. `CHECK_KERNEL()` unchanged. `cudaDeviceSynchronize()` → `cudaStreamSynchronize(stream)` in `sponge.cu`, `bn254_poseidon2.cu`.

### Rust FFI wrappers (~14 files in `src/cuda/`)

Every `extern "C"` declaration and safe wrapper gains `stream: cudaStream_t`.

### Proving pipeline

Intermediate structs (`StackedReductionGpu`, `NttImpl`, `LogupZerocheckGpu`, etc.) store `device_ctx: DeviceContext`. All allocation/copy calls use `_on` variants. Init paths (`ntt.rs`, `batch_ntt_small.rs`) create temporary `DeviceContext` with private streams.

---

## `benchmarks/fields` (5 files)

`.cu` launchers gain `stream` parameter. `lib.rs` FFI wrappers updated. Tests use `DeviceContext`.

## `scripts/gen_stub.py`

Added `using cudaStream_t = void*;` typedef for the `gpu_signature_match` CI ABI checker.

### DIFF
--- a/crates/cuda-backend/cuda/src/bn254_poseidon2.cu
+++ b/crates/cuda-backend/cuda/src/bn254_poseidon2.cu
@@ -455,8 +455,7 @@ __launch_bounds__(BN254_GRIND_BLOCK_SIZE) __global__ void bn254_grind_kernel(
 extern "C" int _init_bn254_poseidon2_rc(
     const uint64_t* initial_rc,
     const uint64_t* partial_rc,
-    const uint64_t* terminal_rc
-) {
+    const uint64_t* terminal_rc, cudaStream_t stream) {
     cudaError_t err;
     err = cudaMemcpyToSymbol(g_initial_rc,  initial_rc,  4 * 3 * 4 * sizeof(uint64_t));
     if (err != cudaSuccess) return (int)err;
@@ -475,8 +474,7 @@ extern "C" int _init_bn254_poseidon2_rc(
 extern "C" int _init_bn254_poseidon2_rc_w2(
     const uint64_t* initial_rc,
     const uint64_t* partial_rc,
-    const uint64_t* terminal_rc
-) {
+    const uint64_t* terminal_rc, cudaStream_t stream) {
     cudaError_t err;
     err = cudaMemcpyToSymbol(g_initial_rc_w2,  initial_rc,  3 * 2 * 4 * sizeof(uint64_t));
     if (err != cudaSuccess) return (int)err;
@@ -492,8 +490,7 @@ extern "C" int _bn254_poseidon2_compressing_row_hashes(
     const Fp*       matrix,
     size_t          width,
     size_t          query_stride,
-    size_t          log_rows_per_query
-) {
+    size_t          log_rows_per_query, cudaStream_t stream) {
     if (log_rows_per_query > 10) {
         return cudaErrorInvalidValue;
     }
@@ -505,10 +502,10 @@ extern "C" int _bn254_poseidon2_compressing_row_hashes(
     size_t shmem_bytes   = shared_stride * sizeof(Bn254Fr);
     auto   height        = query_stride << log_rows_per_query;
 
-    bn254_compressing_row_hashes_kernel<<<grid, block, shmem_bytes>>>(
+    bn254_compressing_row_hashes_kernel<<<grid, block, shmem_bytes, stream>>>(
         out, matrix, width, height, query_stride, log_rows_per_query
     );
-    return CHECK_KERNEL();
+    return CHECK_KERNEL_ON(stream);
 }
 
 extern "C" int _bn254_poseidon2_compressing_row_hashes_ext(
@@ -516,8 +513,7 @@ extern "C" int _bn254_poseidon2_compressing_row_hashes_ext(
     const FpExt*    matrix,
     size_t          width,
     size_t          query_stride,
-    size_t          log_rows_per_query
-) {
+    size_t          log_rows_per_query, cudaStream_t stream) {
     if (log_rows_per_query > 10) {
         return cudaErrorInvalidValue;
     }
@@ -529,20 +525,19 @@ extern "C" int _bn254_poseidon2_compressing_row_hashes_ext(
     size_t shmem_bytes   = shared_stride * sizeof(Bn254Fr);
     auto   height        = query_stride << log_rows_per_query;
 
-    bn254_compressing_row_hashes_ext_kernel<<<grid, block, shmem_bytes>>>(
+    bn254_compressing_row_hashes_ext_kernel<<<grid, block, shmem_bytes, stream>>>(
         out, matrix, width, height, query_stride, log_rows_per_query
     );
-    return CHECK_KERNEL();
+    return CHECK_KERNEL_ON(stream);
 }
 
 extern "C" int _bn254_poseidon2_adjacent_compress_layer(
     bn254_digest_t*       output,
     const bn254_digest_t* prev_layer,
-    size_t                output_size
-) {
+    size_t                output_size, cudaStream_t stream) {
     auto [grid, block] = kernel_launch_params(output_size);
-    bn254_adjacent_compress_layer_kernel<<<grid, block>>>(output, prev_layer, output_size);
-    return CHECK_KERNEL();
+    bn254_adjacent_compress_layer_kernel<<<grid, block, 0, stream>>>(output, prev_layer, output_size);
+    return CHECK_KERNEL_ON(stream);
 }
 
 extern "C" int _bn254_sponge_grind(
@@ -550,8 +545,7 @@ extern "C" int _bn254_sponge_grind(
     uint32_t bits,
     uint32_t min_witness,
     uint32_t max_witness,
-    uint32_t* result
-) {
+    uint32_t* result, cudaStream_t stream) {
     if (bits >= 32 || (uint64_t{1} << bits) >= Fp::P) {
         return cudaErrorInvalidValue;
     }
@@ -559,13 +553,13 @@ extern "C" int _bn254_sponge_grind(
     size_t total_threads = size_t{1} << bits;
     size_t grid_size = div_ceil(total_threads, block_size);
 
-    bn254_grind_kernel<<<grid_size, block_size>>>(init_state, bits, min_witness, max_witness, result);
+    bn254_grind_kernel<<<grid_size, block_size, 0, stream>>>(init_state, bits, min_witness, max_witness, result);
 
     cudaError_t err = cudaGetLastError();
     if (err != cudaSuccess) return (int)err;
 
-    err = cudaDeviceSynchronize();
+    err = cudaStreamSynchronize(stream);
     if (err != cudaSuccess) return (int)err;
 
-    return CHECK_KERNEL();
+    return CHECK_KERNEL_ON(stream);
 }
--- a/crates/cuda-backend/cuda/src/bn254_poseidon2.cu
+++ b/crates/cuda-backend/cuda/src/bn254_poseidon2.cu
@@ -137,7 +137,7 @@ static const int BN254_NUM_F_ELMS    = 8;
 
 /// Row hash for a base-field (Fp / BabyBear) matrix row.
 static __device__ __forceinline__
-Bn254Fr bn254_row_hash(const Fp* matrix, int width, int height, int row) {
+Bn254Fr bn254_row_hash(const Fp* matrix, size_t width, size_t height, size_t row) {
     Bn254Fr state[3];
     state[0] = bn254_zero_init();
     state[1] = bn254_zero_init();
@@ -146,7 +146,7 @@ Bn254Fr bn254_row_hash(const Fp* matrix, int width, int height, int row) {
     uint32_t buf[BN254_BABY_BEAR_RATE];
     int cnt = 0;
 
-    for (int col = 0; col < width; col++) {
+    for (size_t col = 0; col < width; col++) {
         buf[cnt++] = matrix[col * height + row].asUInt32();
         if (cnt == BN254_BABY_BEAR_RATE) {
             state[0] = bn254_pack_base_2_31(buf, BN254_NUM_F_ELMS);
@@ -167,7 +167,7 @@ Bn254Fr bn254_row_hash(const Fp* matrix, int width, int height, int row) {
 
 /// Row hash for an extension-field (FpExt / BinomialExtensionField<BabyBear,4>) matrix row.
 static __device__ __forceinline__
-Bn254Fr bn254_row_hash_ext(const FpExt* matrix, int width, int height, int row) {
+Bn254Fr bn254_row_hash_ext(const FpExt* matrix, size_t width, size_t height, size_t row) {
     Bn254Fr state[3];
     state[0] = bn254_zero_init();
     state[1] = bn254_zero_init();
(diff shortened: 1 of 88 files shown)


gaxiom:
@claude Why do we need CHECK_KERNEL_ON with cudaStreamSynchronize? Why not just to leave CHECK_KERNEL with cudaDeviceSynchronize as it was? The only reason is to have a strict synchronization for debug, isn't it? I think CHECK_KERNEL_ON is a complication that gives almost nothing

crates/cuda-common/src/stream.rs
```diff
@@ -55,67 +75,94 @@ impl CudaStream {
...
+impl StreamGuard {
+    pub fn new(stream: CudaStream) -> Self {
+        Self(Arc::new(stream))
+    }
 
-pub fn current_stream_sync() -> Result<(), CudaError> {
-    check(unsafe { cudaStreamSynchronize(cudaStreamPerThread) })
+    /// Returns `true` if both guards refer to the same underlying `CudaStream`.
+    pub fn is_same_stream(&self, other: &StreamGuard) -> bool {
+        Arc::ptr_eq(&self.0, &other.0)
```
stephenh-axiom-xyz, line 117: Perhaps this should be an implementation of `PartialEq` and `Eq`?

crates/cuda-common/src/d_buffer.rs
```diff
@@ -53,55 +56,76 @@ impl<T> DeviceBuffer<T> {
...
+            len,
+            #[cfg(feature = "debug-cuda-stream")]
+            alloc_stream: std::ptr::null_mut(),
+        }
     }
 
-    /// Allocate device memory for `len` elements of type `T`.
-    pub fn with_capacity(len: usize) -> Self {
+    /// Allocate device memory for `len` elements of type `T` on an explicit stream.
+    pub fn with_capacity_on(len: usize, ctx: &DeviceContext) -> Self {
```
stephenh-axiom-xyz: Do we anticipate doing this on a default stream a lot? Should we add a default `with_capacity`?

crates/cuda-common/src/d_buffer.rs
```diff
@@ -53,55 +56,76 @@ impl<T> DeviceBuffer<T> {
...
         tracing::debug!(
-            "Creating device buffer of size {} (sizeof type = {})",
+            "Creating device buffer of size {} (sizeof type = {}) on stream",
```
stephenh-axiom-xyz: nit: Perhaps we should print out the stream, iirc it has a `Debug` implementation right?

crates/cuda-common/src/d_buffer.rs
```diff
@@ -53,55 +56,76 @@ impl<T> DeviceBuffer<T> {
...
-    pub fn fill_zero(&self) -> Result<(), CudaError> {
+    /// Fills the buffer with zeros on an explicit stream.
+    pub fn fill_zero_on(&self, ctx: &DeviceContext) -> Result<(), CudaError> {
         assert_ne!(self.len, 0, "Empty buffer");
+        #[cfg(feature = "debug-cuda-stream")]
+        debug_assert_eq!(
+            ctx.stream.as_raw(),
+            self.alloc_stream,
+            "fill_zero_on: stream mismatch"
+        );
```
stephenh-axiom-xyz, line 107: Is the expectation that only the stream that allocates a buffer can use it?

crates/cuda-common/src/copy.rs
```diff
@@ -52,23 +47,32 @@ pub unsafe fn cuda_memcpy<const SRC_DEVICE: bool, const DST_DEVICE: bool>(
...
-    fn copy_to(&self, dst: &mut DeviceBuffer<T>) -> Result<(), MemCopyError> {
+    fn copy_to_on(
+        &self,
+        dst: &mut DeviceBuffer<T>,
+        ctx: &DeviceContext,
+    ) -> Result<(), MemCopyError> {
         if self.len() > dst.len() {
             return Err(MemCopyError::SizeMismatch {
-                operation: "copy_to_device",
+                operation: "copy_to_device_on",
```
stephenh-axiom-xyz: nit: mismatching error/function name

crates/cuda-common/src/memory_manager/vm_pool.rs
```diff
@@ -537,18 +539,14 @@ impl VirtualMemoryPool {
...
             ByteSize::b(total_free_size as u64),
-            stream_id
         );
 
-        // Find a best fit unmapped region
         let dst = self.take_unmapped_region(requested)?;
-        // Sentinel value until we have a valid free region pointer from allocation
         let mut allocated_ptr = CUdeviceptr::MAX;
 
-        // Allocate new pages if we don't have enough free regions
```
stephenh-axiom-xyz, line 551: nit: I feel like these comments are still helpful

crates/cuda-common/src/stream.rs
```diff
@@ -55,67 +75,94 @@ impl CudaStream {
...
-#[allow(non_upper_case_globals)]
-pub const cudaStreamPerThread: cudaStream_t = 0x02 as cudaStream_t;
+// ---------------------------------------------------------------------------
+// StreamGuard — keeps a CudaStream alive for allocation records
+// ---------------------------------------------------------------------------
 
-pub type CudaStreamId = u64;
+/// Keeps a `CudaStream` alive for the lifetime of an allocation record.
+#[derive(Clone, Debug)]
+pub struct StreamGuard(Arc<CudaStream>);
```
stephenh-axiom-xyz, line 107: This doesn't really seem like it does much. Based on the description and how it's used above, this could probably just be an alias for `Arc<CudaStream>`?

crates/cuda-common/src/memory_manager/mod.rs
```diff
@@ -98,25 +108,28 @@ impl MemoryManager {
         Ok(ptr)
     }
 
+    /// Two-stage free: first resolves the record under the lock, returning the
+    /// `StreamGuard` that must be dropped AFTER the lock is released.
+    ///
     /// # Safety
```
stephenh-axiom-xyz, line 114: nit: add a safety comment stipulating that the lock must be acquired before calling `d_free_under_lock`

crates/cuda-backend/cuda/src/logup_zerocheck/batch_mle.cu
```diff
@@ -385,16 +387,15 @@ extern "C" int _zerocheck_batch_eval_mle(
     uint32_t num_blocks,
     uint32_t num_x,
     uint32_t num_airs,
-    uint32_t threads_per_block
-) {
+    uint32_t threads_per_block, cudaStream_t stream) {
```
stephenh-axiom-xyz: nit: formatting (here and in many instances below)

crates/cuda-backend/cuda/src/bn254_poseidon2.cu
```diff
@@ -550,21 +545,20 @@ extern "C" int _bn254_sponge_grind(
...
     size_t grid_size = div_ceil(total_threads, block_size);
 
-    bn254_grind_kernel<<<grid_size, block_size>>>(init_state, bits, min_witness, max_witness, result);
+    bn254_grind_kernel<<<grid_size, block_size, 0, stream>>>(init_state, bits, min_witness, max_witness, result);
 
     cudaError_t err = cudaGetLastError();
     if (err != cudaSuccess) return (int)err;
 
-    err = cudaDeviceSynchronize();
+    err = cudaStreamSynchronize(stream);
```
stephenh-axiom-xyz, line 561: Unrelated to the PR, but why do we sync here? This feels like it should be a debug

crates/cuda-backend/src/logup_zerocheck/batch_mle.rs
```diff
@@ -117,6 +126,7 @@ pub(crate) struct ZerocheckMleBatchBuilder<'a> {
     air_offsets: DeviceBuffer<u32>,
     threads_per_block: u32,
     _intermediates_keepalive: Vec<DeviceBuffer<EF>>,
+    ctx: DeviceContext,
```
stephenh-axiom-xyz: nit: should we be cloning `DeviceContext`? Given that this struct is already generic in `'a`, perhaps this could be reference? Not sure if this breaks smth though

crates/cuda-backend/src/logup_zerocheck/batch_mle_monomial.rs
```diff
@@ -285,6 +293,7 @@ pub(crate) struct ZerocheckMonomialParYBatch<'a> {
     air_offsets: DeviceBuffer<u32>,
     num_blocks: u32,
     chunk_size: u32,
+    ctx: DeviceContext,
```
stephenh-axiom-xyz: Same comment on cloning `ctx` as above

crates/cuda-backend/src/device.rs
```diff
@@ -1,14 +1,16 @@
...
     device_info::get_sm_count,
 };
 
+/// Pure device configuration — no stream, no CUDA runtime state.
+/// Renamed from the old `GpuDevice`.
 #[derive(Clone, Getters, CopyGetters, MutGetters)]
-pub struct GpuDevice {
+pub struct GpuDeviceConfig {
     #[getset(get = "pub")]
     pub(crate) config: SystemParams,
```
stephenh-axiom-xyz: If we're changing the constructor to `params: SystemParams` we should probably change it here too

crates/cuda-backend/src/device.rs
```diff
@@ -25,34 +27,66 @@ pub struct GpuProverConfig {
...
             id,
             sm_count,
-        }
+        };
+
+        Ok(Self { config, ctx })
+    }
+
+    // Delegate accessors to inner config
+    pub fn config(&self) -> &SystemParams {
```
stephenh-axiom-xyz: Related to comment above, we should probably rename this to `params`

stephenh-axiom-xyz (commented):
Mostly LGTM, some comments. Also, on `DeviceContext`:

- We use `*Context` a lot (probably more than we should), so except in places where it's extremely obvious what `ctx` refers to we should probably use `device_ctx`
- I'm not sure if we should be cloning `DeviceContext` and passing it to intermediate structs, I'd much rather either (a) pass it into functions directly or at least (b) store the reference

Please take a look at the Claude-reported issues too

crates/cuda-common/src/d_buffer.rs
gaxiom: The entire point of the migration is to eliminate implicit stream usage. We probably can rename every `_on` function back to original, but each call should explicitly declare which stream it belongs to

crates/cuda-common/src/d_buffer.rs
gaxiom: Yes. Somewhere in the future we might want to do multi-streaming - then we should handle it with events. Current impl expects that

crates/cuda-common/src/stream.rs
gaxiom: Thought about it at the beginning, but switched to type https://hackmd.io/CGcmUvNUSYOLtYXQJuDvFw?view#2-StreamGuard-%E2%80%94-Newtype-for-Ref-Counted-Stream-Ownership

crates/cuda-backend/cuda/src/bn254_poseidon2.cu
gaxiom: The grind kernel returns the PoW result in device memory — the sync is needed so the host can read it. Making it debug-only would cause incorrect results.

crates/cuda-backend/src/logup_zerocheck/batch_mle.rs
gaxiom: DeviceContext clone is cheap - just `(u32, Arc<cudaStream>)`. I don't like the idea of another borrow from a different source

stephenh-axiom-xyz (approved):
LGTM after the renaming we discussed offline

PR #318, opened 2026-04-09
feat: Add synthetic stark-backend benchmark

georgwiese:
Configurable synthetic benchmark for measuring proving time vs number of AIRs, columns, constraints, and bus interactions. Uses zero witness with boolean constraints and self-canceling send/receive interactions.

For a tiny example:
```
cargo run -p openvm-benchmark-proving --release -- \
  --num-airs 4 --cols-per-air 10 --log-rows-per-air 18 --log-stacked-height 18
```

With CUDA:
```
cargo run -p openvm-benchmark-proving --release --features cuda -- \
  --num-airs 4 --cols-per-air 10 --log-rows-per-air 18 --log-stacked-height 18
```

It works roughly as follows:
- By default, all columns are boolean-constrained (i.e., there is as many constraints as columns)
- By default, there is 1 bus interaction for every 4 columns
- The number of rows per AIR is configurable
- An all-zero trace is passed to the prover

Cherry-picked from https://github.com/powdr-labs/stark-backend/pull/24

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = [
     "crates/cuda-builder",
     "crates/cuda-backend",
     "benchmarks/fields",
-    "benchmarks/air-complexity",
+    "benchmarks/proving",
 ]
 # Only these build by default
 default-members = [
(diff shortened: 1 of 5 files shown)


jonathanpwang:
@claude can you fix the lint by running `cargo +nightly fmt --all`

jonathanpwang (approved):
LGTM - very nice idea

PR #322, opened 2026-04-11
fix: soundness calculator list size terms

jonathanpwang:
The WHIR paper has:
image
for $i > 0$. 

In our soundness calculator, we iterate over `round in 0..` and so our handling should match up to `i = round + 1`.
We have a few discrepancies that are resolved in this PR:
* fold term needed the $d \cdot \ell / |F|$ term
* OOD error and shift error both reference $\ell_{i,0}$ which should be the list size of the _next_ round, since the query error is $\delta_{i-1}$ for the "current" round.

The following is only for convenience to match the paper's theorem statement:
* We update the query soundness to use the weaker Guraswami-Sudan bound, although we believe the strong bound still holds if one propagates a more careful analysis of the proof. To simplify and align with the paper and other calculators that are derived from the paper (e.g., soundcalc), we use the looser bound. This may cause a very small increase to the number of queries.
* To better match the paper, I also removed the `subround == 0` branching since using `d = 3` everywhere is simpler and we don't need the finer analysis.

We adjust the root params so that it still meets the 100 bits security threshold.

ref INT-6720

### DIFF
--- a/crates/stark-backend/src/soundness.rs
+++ b/crates/stark-backend/src/soundness.rs
@@ -361,7 +361,7 @@ impl SoundnessCalculator {
             // sub-round
             let mut log2_list_size: Option<f64> = None;
 
-            for sub_round in 0..k_whir {
+            for _ in 0..k_whir {
                 current_log_degree -= 1;
 
                 let prox_gaps = Self::whir_proximity_gap_security(
@@ -381,7 +381,6 @@ impl SoundnessCalculator {
 
                 let sumcheck_bits = Self::whir_sumcheck_security(
                     challenge_field_bits,
-                    sub_round,
                     whir.folding_pow_bits,
                     log2_list_size.unwrap(),
                 );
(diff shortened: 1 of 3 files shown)


crates/stark-backend/src/soundness.rs
```diff
@@ -677,15 +677,16 @@ impl SoundnessCalculator {
 
     /// Computes WHIR sumcheck security bits for a sub-round.
     ///
-    /// Sumcheck error is 2/|F| for sub_round 0 and 3/|F| for sub_round > 0.
-    /// Security bits = |F_ext| - log₂(degree) + folding_pow_bits
+    /// Sumcheck error is d * ℓ / |F|, d^*:= 1 + deg_Z(w0) + max_i deg_{X_i}(w0) and d :=
+    /// max{d^*,3}. Security bits = |F_ext| - log₂(degree) + folding_pow_bits
```
zlangley: Security bits needs updating.

zlangley (approved):

zlangley (approved):
LGTM

PR #326, opened 2026-04-23
fix: harden proof wrt q0_claim

zlangley:
When there are no interactions, the verifier now enforces that q0_claim must be one.

### DIFF
--- a/crates/stark-backend/tests/fractional_sumcheck_gkr.rs
+++ b/crates/stark-backend/tests/fractional_sumcheck_gkr.rs
@@ -247,27 +247,3 @@ fn test_gkr_mixed_fractions() -> eyre::Result<()> {
     assert_ne!(denom_claim, EF::ZERO);
     Ok(())
 }
-
-#[test]
-fn test_gkr_empty_case() -> eyre::Result<()> {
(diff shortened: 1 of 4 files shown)


stephenh-axiom-xyz (approved):
LGTM, do we need to update zkSecurity on thsi?

jonathanpwang (changes requested):
Wait this was not what was described in Linear - can't we just delete this if statement entirely? `verify_gkr` already has a branch to handle this

crates/stark-backend/tests/fractional_sumcheck_gkr.rs
```diff
@@ -247,27 +247,3 @@ fn test_gkr_mixed_fractions() -> eyre::Result<()> {
     assert_ne!(denom_claim, EF::ZERO);
     Ok(())
 }
-
-#[test]
-fn test_gkr_empty_case() -> eyre::Result<()> {
```
jonathanpwang, line 252: do we have a test for this case for the batch constraints verify?

jonathanpwang (approved):
LGTM but make sure we test this case in verify batch constraints

PR #330, opened 2026-04-28
chore: set target security bits to 128

stephenh-axiom-xyz:
Towards INT-7602.

Sets target bits of security to 128, and correspondingly reduces proof-of-work bit params to slightly optimize.

### DIFF
--- a/crates/stark-sdk/src/config/mod.rs
+++ b/crates/stark-sdk/src/config/mod.rs
@@ -54,8 +54,8 @@ pub fn app_params_with_128_bits_security(log_stacked_height: usize) -> SystemPar
         log_stacked_height.saturating_sub(DEFAULT_APP_L_SKIP), // n_stack
         2048,                                                  // w_stack
         WHIR_MAX_LOG_FINAL_POLY_LEN,
-        5,  // folding pow
-        15, // mu pow
+        0,  // folding pow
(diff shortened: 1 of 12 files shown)


crates/stark-sdk/src/config/mod.rs
```diff
@@ -136,17 +136,17 @@ pub fn internal_params_with_100_bits_security() -> SystemParams {
...
+pub fn root_params_with_128_bits_security() -> SystemParams {
     SystemParams::new(
         DEFAULT_ROOT_LOG_BLOWUP,
         2,  // l_skip
         18, // n_stack
         19, // w_stack
         WHIR_MAX_LOG_FINAL_POLY_LEN,
-        20, // folding pow
-        20, // mu pow
+        12, // folding pow
```
jpw-axiom, line 146: interesting this could be lowered, can we go back to `n_stack = 17` and `w_stack = 10` (we can tune this later)

jpw-axiom (approved):

crates/stark-sdk/src/config/mod.rs
stephenh-axiom-xyz: I don't think we can lower `w_stack` without increasing `n_stack` or `l_skip` given the number of cells the root verifier currently has

PR #331, opened 2026-04-29
perf(cuda): virtualize GKR input padding

jonathanpwang:
## Summary

This PR virtualizes CUDA fractional-sumcheck GKR input padding. The GKR input eval buffer is allocated at the real interaction length `R` instead of the logical power-of-two length `L`, and the existing fused CUDA kernel structure is preserved by recovering virtual padding values on demand.

It also keeps the safety fixes found during review and reth benchmarking:

- compact virtual final folds avoid in-place mutation when virtual reads can alias compact source slots;
- the precompute-M work buffer is sized for the largest virtual last-round multifold output;
- `AGENTS.md` now requires `cargo +nightly fmt` and crate-scoped `cargo clippy ... -D warnings` before committing or opening/updating a PR.

### Rust orchestration

- `FractionalInputSize { real_len, logical_len }` now carries both the compact real input length and the logical dense domain length into `fractional_sumcheck_gpu`.
- `log_gkr_input_evals` allocates only `real_len` `Frac<EF>` entries for the input evals and no longer materializes padding entries up to `logical_len`.
- `fractional_sumcheck_gpu` validates `real_len <= logical_len`, `logical_len.is_power_of_two()`, and `logical_len / 2 <= real_len`, then threads both lengths through tree build, revert, fold, fused compute+fold, precompute-M, and multifold paths.
- The final compact FoldEval fold uses a tiny out-of-place scratch when `source_real_len < source_logical_len`; this prevents an in-place write from overwriting a compact source slot that another lane still needs to recover virtually.
- The precompute-M work buffer formula was corrected for the first virtual last-round multifold. The first precompute-M multifold folds `w + 1` variables, so its largest output is `L >> (1 + w)`, not `L >> (2 + w)`.

### CUDA kernels and FFI

- Existing GKR kernels were edited in place; no duplicate kernels were introduced.
- The CUDA FFI wrappers now pass `real_len`, `logical_len`, `dst_real_len`, `dst_logical_len`, and `alpha` where needed.
- `virtual_node_value` reconstructs logical entries that are not physically stored:
  - all-padding subtrees return `(0, alpha^subtree_len)`;
  - compact real spill entries are read from their physical compact slot;
  - physically present entries are read directly.
- `virtual_node_store` writes only entries that have compact physical storage and skips all-padding outputs.
- The following existing kernels now use the virtual read/store helpers where applicable:
  - tree layer build and fused two-layer build;
  - fused bit-reversal + K=2 tree build;
  - first-round revert + compute;
  - fold and fused compute+fold;
  - precompute-M build;
  - multifold.
- Dense in-place folds remain in-place. Compact virtual folds must be out-of-place when virtual reads can map into the destination range; the Rust wrapper now has a debug assertion documenting this invariant.

### Tests and tooling

- Added dense-vs-virtual fractional GKR proof-equivalence tests, including nontrivial `alpha` and forced precompute-M coverage.
- Updated `bench.rs` and CUDA NTT helper wiring as needed for the new compact/virtual path.

## Example Walkthrough

Take `real_len = R = 9` and `logical_len = L = 16`.

Before this PR, the input evals buffer had 16 entries: 9 real entries plus 7 materialized padding entries. With this PR, the input evals buffer has only entries `0..8`.

When a kernel asks for logical dense entry `dense_idx`, it computes:

```text
subtree_len = logical_len / active_size
start = bit_reverse(dense_idx, log2(active_size)) * subtree_len
```

Then:

- if `start >= R`, the whole subtree is virtual padding, so the value is recovered as `(0, alpha^subtree_len)` and no memory is read;
- if `dense_idx < R`, the value is physically present at `layer[dense_idx]`;
- otherwise, the logical slot is a compact-layout hole whose real value is stored at `layer[start]`.

For example with `active_size = 16`:

- `dense_idx = 15` gives `start = 15 >= 9`, so the value is virtual padding: `(0, alpha)`.
- `dense_idx = 10` gives `start = 5 < 9`, so the logical slot is recovered from compact physical entry `layer[5]` instead of reading missing entry `layer[10]`.

Stores use the same mapping. Padding-only outputs are skipped, and mixed real/padding outputs are stored in their compact physical slot. This lets the GKR tree/fold kernels behave as if the dense `L`-sized buffer existed without actually allocating or filling it.

The important in-place caveat is that compact virtual reads can map a later logical source entry into an early physical slot. If a fold writes that early slot before every lane has consumed it, the dense-vs-virtual semantics diverge. The final FoldEval compact fold therefore writes to a small scratch buffer instead of mutating the compact source in place.

## Fractional GKR Peak Memory Formula

Let:

```text
R = real interaction count
L = logical GKR length = 1 << (l_skip + n_logup), with L / 2 <= R <= L
w = GKR_WINDOW_SIZE = 3
N_min = GKR_WINDOW_DEFAULT_MIN_N = 22
S_frac = sizeof(Frac<EF>)
S_ef = sizeof(EF) = S_frac / 2
```

The dominant `Frac<EF>` allocations inside fractional-sumcheck GKR are now:

```text
layer = R

work_buffer =
  FoldEval:    L / 4
  PrecomputeM: max(L / 16, 2^N_min)

final_fold_buffer <= 2
```

So the `Frac<EF>` contribution is:

```text
FoldEval peak Frac entries    = R + L / 4 + O(1)
PrecomputeM peak Frac entries = R + max(L / 16, 2^22) + O(1)
```

Because virtual padding requires `L / 2 <= R`, the variable precompute-M work-buffer term satisfies:

```text
L / 16 <= R / 8
```

The `2^22` minimum is inherited from the existing precompute-M fallback sizing and can dominate smaller `L`; for the reth benchmark run, this minimum already dominated on `develop-v2`, so the corrected `L / 16` bound did not increase the allocated work buffer for that run.

Precompute-M also uses EF-only auxiliary buffers:

```text
m_total        = 2^(2w) EF entries
m_partial      = ceil(2^(rem_n - w) / tail_tile) * 2^(2w) EF entries, max over active windows
eq_prefix      = 2^w EF entries
eq_suffix      = 2^w EF entries
eq_r_window    = reused tmp_block_sums storage, no separate allocation
```

Thus the precompute-M EF auxiliary contribution is:

```text
M_precompute_EF =
  2^(2w)
  + max_window(ceil(2^(rem_n - w) / tail_tile) * 2^(2w))
  + 2^(w + 1)
```

In bytes, the fractional-sumcheck GKR contribution from the new compact layer/work-buffer scheme is:

```text
Peak bytes ~= S_frac * (R + work_buffer + final_fold_buffer)
           + S_ef * M_precompute_EF
           + existing eq/tmp round buffers
```

The key change is that the main GKR input/layer buffer is `R`, not `L`; the PR does not allocate an `L`-sized dense padded input buffer.

## Validation

- `cargo +nightly fmt -- --check`
- `cargo check -p openvm-cuda-backend`
- `cargo test -p openvm-cuda-backend test_virtual_input_padding_matches_dense_padding -- --nocapture --test-threads=1`
- `cargo clippy -p openvm-cuda-backend --all-targets --tests -- -D warnings`
- Patched `../openvm-eth` to use this local `stark-backend`, rebuilt `openvm-reth-benchmark` with `cuda,aot`, and ran the reth benchmark in `prove-stark` + CUDA mode.

Reth benchmark result:

```text
block: 23992138
segments: 149
peak GPU memory: 15.66 GB
proof size: 315223 bytes
compressed proof size: 279121 bytes
exit code: 0
```

### DIFF
--- a/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
+++ b/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
@@ -1744,20 +1744,19 @@ T bit_rev(T i, unsigned int nbits)
 //   - Extra `alpha` parameter: adds alpha to every denominator after bit-reversing
 //   - Two GKR tree layers (K=2) are computed inside shmem before writing to global mem,
 //     eliminating two separate kernel passes
-//   - __syncwarp() is used unconditionally (original conditionally uses __syncthreads()
-//     when Z_COUNT > WARP_SIZE, which is never true for Z_COUNT=8)
+//   - Masked __syncwarp(subgroup_mask) is used for each 8-lane Z-tile (matching
+//     the original Z_COUNT <= WARP_SIZE path)
 //
 // Output layout for each Z-tile written at base_rev (i*step + base_rev, i=0..7):
 //   [0,1]  : layer-1 results  (tree level 2)
 //   [2,3]  : layer-0 results preserved (tree level 1 right half, for revert of layer 1)
 //   [4..7] : original+alpha preserved (leaves, for revert of layer 0)
 //
-// __launch_bounds__(256): not a correctness constraint — __syncwarp() is safe for any
-// multiple-of-32 bsize because each gid group is Z_COUNT=8 threads (always within one
-// warp) and different warps operate on independent shmem tiles. bsize=256 was chosen after
-// benchmarking: 2× faster than bsize=32 in isolation (1525 vs 754 GB/s at lg=24 on RTX
-// 5090). It requires 64KB dynamic shmem per block, so cudaFuncSetAttribute is used in the
-// launcher — but only once (static local), paid on the first call and never again.
+// __launch_bounds__(256): not a correctness constraint. Sync is per 8-lane Z-tile
+// using subgroup masks because branch predicates are uniform per tile, not per whole
+// warp. bsize=256 was chosen after benchmarking: 2x faster than bsize=32 in isolation
+// (1525 vs 754 GB/s at lg=24 on RTX 5090). It requires 64KB dynamic shmem per block,
+// so cudaFuncSetAttribute is used in the launcher, paid once on the first call.
 __launch_bounds__(256) __global__
 void bit_rev_frac_build_k2_kernel(
     FracExt* inout, uint32_t real_len, uint32_t lg_domain_size, FpExt alpha)
@@ -1772,6 +1771,9 @@ void bit_rev_frac_build_k2_kernel(
     uint32_t gid = threadIdx.x / Z_COUNT;
     uint32_t idx = threadIdx.x % Z_COUNT;
     uint32_t rev = bit_rev(idx, LG_Z_COUNT);
+    uint32_t lane = threadIdx.x & (WARP_SIZE - 1);
+    uint32_t subgroup_base = lane - idx;
+    unsigned subgroup_mask = ((1u << Z_COUNT) - 1u) << subgroup_base;
 
     index_t domain_size = (index_t)1 << lg_domain_size;
     index_t step = (index_t)1 << (lg_domain_size - LG_Z_COUNT);
@@ -1803,7 +1805,7 @@ void bit_rev_frac_build_k2_kernel(
             }
         }
 
-        __syncwarp();
+        __syncwarp(subgroup_mask);
 
         // Apply alpha + tree layers 0 and 1 to transposed tile, then write to base_rev
         {
@@ -1841,13 +1843,13 @@ void bit_rev_frac_build_k2_kernel(
         if (group_idx == group_rev)
             continue;
 
-        __syncwarp();
+        __syncwarp(subgroup_mask);
 
         #pragma unroll
         for (uint32_t i = 0; i < Z_COUNT; i++)
             xchg[gid][i][rev] = regs[i];
 
-        __syncwarp();
+        __syncwarp(subgroup_mask);
 
         // Apply alpha + tree layers 0 and 1 to transposed tile, then write to base_idx
         {
--- a/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
+++ b/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
@@ -24,10 +24,66 @@ constexpr int GKR_SP_DEG = 2;
 // ============================================================================
 // KERNELS
 // ============================================================================
+__device__ __forceinline__ uint32_t lg_pow2(uint32_t x) {
+    return 31u - __clz(x);
+}
+
+__device__ __forceinline__ FpExt virtual_padding_q(FpExt alpha, uint32_t subtree_len) {
+    FpExt q = alpha;
+    for (uint32_t len = subtree_len; len > 1; len >>= 1) {
+        q = q * q;
+    }
+    return q;
+}
+
+__device__ __forceinline__ FracExt virtual_node_value(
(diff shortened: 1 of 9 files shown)


crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
```diff
@@ -24,43 +24,143 @@ constexpr int GKR_SP_DEG = 2;
...
+    const bool virtual_mode = real_len < logical_len;
+    if constexpr (apply_alpha) {
+        if (virtual_mode && layer_size == logical_len) {
+            const uint32_t parent_active_size = half;
+            const uint32_t parent_start =
+                rev_len(idx, lg_pow2(parent_active_size)) << 1;
+
+            FracExt lhs_val = parent_start < real_len
+                ? layer[parent_start]
+                : FracExt{FpExt(Fp::zero()), FpExt(Fp::zero())};
```
gaxiom: Maybe cleaner to define FracExt::zero()?

crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
```diff
@@ -24,43 +24,143 @@ constexpr int GKR_SP_DEG = 2;
...
 // KERNELS
 // ============================================================================
+__device__ __forceinline__ uint32_t lg_pow2(uint32_t x) {
+    return 31u - __clz(x);
+}
+
+__device__ __forceinline__ FpExt virtual_padding_q(FpExt alpha, uint32_t subtree_len) {
+    FpExt q = alpha;
+    for (uint32_t len = subtree_len; len > 1; len >>= 1) {
+        q = q * q;
```
gaxiom: Not sure but `q *= q;` can be slightly more optimal

crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
```diff
@@ -24,43 +24,143 @@ constexpr int GKR_SP_DEG = 2;
...
+        : layer[idx | half];
+
     if constexpr (apply_alpha) {
         // When applying alpha, we need to modify both operands before combining
         // Read both values, apply alpha to denominators, then combine
-        FracExt lhs_val = layer[idx];
-        FracExt rhs_val = layer[idx | half];
-        lhs_val.q = lhs_val.q + alpha;
-        rhs_val.q = rhs_val.q + alpha;
+        lhs.q = lhs.q + alpha;
```
gaxiom: Same for `+=`

crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
```diff
@@ -24,43 +24,143 @@ constexpr int GKR_SP_DEG = 2;
...
+
+            FracExt lhs_val = parent_start < real_len
+                ? layer[parent_start]
+                : FracExt{FpExt(Fp::zero()), FpExt(Fp::zero())};
+            FracExt rhs_val = parent_start + 1 < real_len
+                ? layer[parent_start + 1]
+                : FracExt{FpExt(Fp::zero()), FpExt(Fp::zero())};
+            FracExt rhs_leaf = rhs_val;
+            bool has_rhs_leaf = parent_start + 1 < real_len;
+            __syncthreads();
```
gaxiom, line 114: Combination of `__syncthreads();` and `if (idx >= half) return;` works fine only if `half % 32 == 0` always (one WARP) - otherwise potential undefined behaviour

gaxiom (approved):
LGTM. Nothing critical, but synthreads worth fixing

PR #334, opened 2026-04-30
perf: partial batch GKR input eval CUDA kernels across AIRs

jonathanpwang:
## Summary
- Launch a single CUDA kernel per batch instead of one kernel per AIR for GKR input eval, increasing GPU parallelism
- Use partial batching controlled by a memory budget derived from the fractional-sumcheck GKR work-buffer peak, so batched intermediates never push total peak memory beyond what the non-batched code already uses
- Add `FractionalInputSize::peak_work_buffer_bytes()` co-located with the sumcheck work-buffer sizing so future changes to fractional sumcheck GKR stay in sync

## Changes
- `cuda/src/logup_zerocheck/gkr_input.cu` — batch kernel `evaluate_interactions_gkr_batch_kernel` reusing the existing DAG evaluation helper
- `src/cuda/logup_zerocheck.rs` — `GkrInputCtx` FFI struct and wrapper functions
- `src/logup_zerocheck/gkr_input.rs` — rewritten `log_gkr_input_evals` with two-phase greedy batching and per-AIR buffer keepalive
- `src/logup_zerocheck/fractional.rs` — `peak_work_buffer_bytes()` on `FractionalInputSize`
- `src/logup_zerocheck/mod.rs` — compute and pass the budget to `log_gkr_input_evals`

### DIFF
--- a/crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
+++ b/crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
@@ -384,10 +384,7 @@ __global__ void evaluate_interactions_gkr_batch_kernel(
 }
 
 // Size helper: returns number of FpExt elements needed for intermediates per AIR.
-extern "C" size_t _gkr_input_batch_intermediates_buffer_size(
-    uint32_t buffer_size,
-    cudaStream_t stream
-) {
+extern "C" size_t _gkr_input_batch_intermediates_buffer_size(uint32_t buffer_size) {
     return (size_t)TASK_SIZE * buffer_size;
 }
 
--- a/crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
+++ b/crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
@@ -1,3 +1,4 @@
+#include "block_ctx.cuh"
 #include "codec.cuh"
 #include "fp.h"
 #include "fpext.h"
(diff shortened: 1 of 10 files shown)


crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
```diff
@@ -263,4 +263,142 @@ extern "C" int _logup_gkr_input_eval(
...
+                    size_t out_idx = (interaction_idx * ctx.height + row) * 2;
+                    reinterpret_cast<FpExt *>(ctx.d_fracs)[out_idx + (pair_idx & 1)] = result;
+                }
+            }
+        }
+    }
+}
+
+// Size helper: returns number of FpExt elements needed for intermediates per AIR.
+extern "C" size_t _gkr_input_batch_intermediates_buffer_size(
```
gaxiom: Why we need this?

crates/cuda-backend/src/logup_zerocheck/fractional.rs
```diff
@@ -56,6 +56,43 @@ impl FractionalInputSize {
...
+        let fold_eval = (self.logical_len / 4) * s_frac;
+
+        let s_ef = std::mem::size_of::<EF>();
+        let w = GKR_WINDOW_SIZE;
+        let precompute_f =
+            (self.logical_len >> (1 + w)).max(1 << GKR_WINDOW_DEFAULT_MIN_N) * s_frac;
+        // Conservative estimate for M_precompute_EF: m_total + m_partial bound + eq_prefix/suffix.
+        // In practice the max_window term is ceil(2^(rem_n - w) / tail_tile) * 2^(2w) EF elems;
+        // use 2 * 2^(2w) as a safe floor (tail_tile >= 1, rem_n bounded by total_rounds).
+        let precompute_ef = (2 * (1 << (2 * w)) + (1 << (w + 1))) * s_ef;
```
gaxiom: `(1 << (2w + 1)) + (1 << (w + 1))` OR `2 * (1 << 2w + 1 << w)`
Just a little messy to combine it that way, don't you think?

gaxiom (approved):
Seems ok if all tests passed

crates/cuda-backend/src/logup_zerocheck/fractional.rs
jonathanpwang: I like the first one

crates/cuda-backend/cuda/src/logup_zerocheck/gkr_input.cu
jonathanpwang: You mean the function? It's following our pattern that any temp buffer sizing is owned by the .cu since it might depend on launch params

PR #335, opened 2026-05-04
feat: remove Permutation and Exposed from Entry enum

stephenh-axiom-xyz:
Resolves INT-7707

## Overview

`SymbolicVariable<F>::entry: Entry` had six variants. The prover/verifier no longer use a permutation trace, and exposed-after-challenge values are unused — but the variants and their plumbing were still present. This PR retires `Entry::Permutation` and `Entry::Exposed` everywhere they appeared (Rust + CUDA), and deletes the dead `SymbolicRapBuilder` infrastructure that existed only to construct them. The `TraceWidth.after_challenge` field is also removed since it was always empty after this cleanup. `Entry::Challenge` is kept.

After the change, `Entry` has four variants: `Preprocessed`, `Main`, `Public`, `Challenge`. `Challenge` is no longer constructible through `SymbolicRapBuilder`; the only producer in-tree is `cuda-backend/src/pkey.rs` (unchanged in this PR).

There are no behavior changes — every removed code path was either dead or `unreachable!()` in practice.

Reviewer flow: read in this order. Each section is self-contained.

---

## 1. The Entry enum (`stark-backend`)

**File:** `crates/stark-backend/src/air_builders/symbolic/symbolic_variable.rs`

`Permutation { offset }` and `Exposed` are removed from the `Entry` enum. The `offset()`, `Entry::rotate()`, and `SymbolicVariable::degree_multiple()` match arms are simplified accordingly. `Challenge` stays.

This is the only "real" change — everything else in the PR is making the rest of the workspace consistent with this.

---

## 2. `SymbolicRapBuilder` cleanup (`stark-backend`)

**File:** `crates/stark-backend/src/air_builders/symbolic/mod.rs`

The builder previously constructed `Permutation` / `Challenge` / `Exposed` symbolic variables for each AIR. Those constructions and everything that fed them are deleted because `permutation()` (the only consumer of the `after_challenge` matrix) no longer has an `Entry` to use.

Removed:
- Fields: `after_challenge`, `challenges`, `exposed_values_after_challenge`.
- Helpers: `new_after_challenge`, `new_challenges`, `new_exposed_values_after_challenge`.
- Deprecated accessors: `num_challenges_to_sample()`, `num_exposed_values_after_challenge()`.
- `impl PermutationAirBuilder for SymbolicRapBuilder<F>` (and the `PermutationAirBuilder` import).
- `Permutation`/`Exposed` arms in the `LocalOnlyChecker::check_var` match.

API change: `get_symbolic_builder` and `SymbolicRapBuilder::new` lose their `num_challenges_to_sample` and `num_exposed_values_after_challenge` `&[usize]` parameters — both were always passed `&[]` from the only call site (`AirKeygenBuilder::get_symbolic_builder`, updated below).

`width()` now just clones `trace_width` (the `after_challenge` field on `TraceWidth` is removed in section 3).

---

## 3. `TraceWidth.after_challenge` removal (`stark-backend`)

**Files:** `crates/stark-backend/src/keygen/types.rs`, `crates/stark-backend/src/keygen/mod.rs`, `crates/stark-backend/src/prover/metrics.rs`

`TraceWidth.after_challenge: Vec<usize>` is removed — it was always `vec![]` in the only construction site, and `assert!(...is_empty())` enforced it at keygen time. Knock-on changes:

- `TraceWidth::total_width(&self, ext_degree: usize) -> usize` — kept, but `after_challenge` term and the `ext_degree` parameter are gone (no callers in this repo). External callers (e.g. openvm) need to drop the `ext_degree` arg.
- `keygen/mod.rs` — `assert!(vparams.width.after_challenge.is_empty())` and the `after_challenge: vec![]` initializer are deleted.
- `prover/metrics.rs` — the `TraceCells.after_challenge` field, the `interactions × ext_degree × height` computation that fed it, the `perm_trace_cells` info log, the `Display` impl's `Perm Cols = …` column, and the `perm_cols` metrics counter are all deleted. `total_cells` no longer includes that contribution.

## 4. Match-arm fallout in `stark-backend`

These sites previously handled `Permutation`/`Challenge`/`Exposed` together in an `unreachable!` arm. With the variants gone, the match is narrower; `Challenge` is now an explicit (still `unreachable!` in these paths) arm.

- `crates/stark-backend/src/keygen/mod.rs`
  - `find_unused_vars`: `Entry::Challenge | Exposed | Permutation` → `Entry::Challenge => unreachable!()`.
  - Updated `get_symbolic_builder(...)` call to drop the two `&[]` args.
- `crates/stark-backend/src/interaction/debug.rs`: `_ => unreachable!("There should be no after challenge variables")` → explicit `Entry::Challenge => unreachable!("There should be no challenge variables")`.
- `crates/stark-backend/src/prover/logup_zerocheck/evaluator.rs`: same pattern, message updated to `"challenge not supported"`.
- `crates/stark-backend/src/prover/logup_zerocheck/cpu.rs`: `_ => Err(AfterChallengeNotSupported)` → explicit `Entry::Challenge => Err(ChallengeNotSupported)`.
- `crates/stark-backend/src/prover/error.rs`: `LogupZerocheckError::AfterChallengeNotSupported` → `ChallengeNotSupported`; error message text updated to `"challenge not supported"`.

---

## 5. Lean renderer (`stark-backend`)

**File:** `crates/stark-backend/src/lean/render.rs`

The Lean placeholder/leaf renderers had explicit cases for all six variants (with `Permutation` already an `unreachable!`). Those `unreachable!` and the `Exposed` rendering are dropped; `Challenge` rendering is unchanged. `get_entry_type_id` still returns `3` for `Challenge` (the gap from removing `Permutation` is intentional — the type IDs aren't part of any persisted format here).

---

## 6. `cpu-backend`

**File:** `crates/cpu-backend/src/logup_zerocheck/mod.rs`

Three `_ => unreachable!("after_challenge not supported")` messages updated to `"challenge not supported"`. No structural changes — these arms remain catch-alls. (Stark-backend's analogous sites were converted to explicit `Entry::Challenge` arms; cpu-backend was left as `_ =>` to minimize churn. Worth a one-line opinion if you'd prefer them aligned.)

---

## 7. `cuda-backend`

Two Rust files and one CUDA header. The Rust codec and the C++ codec enum must stay in lockstep.

**`crates/cuda-backend/src/monomial.rs`** — `PackedVar::from_symbolic_var` match arm narrowed from `Permutation | Challenge | Exposed` to just `Challenge` (still `panic!("unsupported symbolic entry in zerocheck monomial extraction")`).

**`crates/cuda-backend/src/logup_zerocheck/rules/codec.rs`** — encoder/decoder for `Entry`/`SymbolicVariable`/`Source`. `PERMUTATION` and `EXPOSED` constants and their encode/decode arms are removed. The remaining tags are **renumbered**:

| Variant       | Before | After |
|---------------|--------|-------|
| PREPROCESSED  | 0      | 0     |
| MAIN          | 1      | 1     |
| PERMUTATION   | 2      | —     |
| PUBLIC        | 3      | 2     |
| CHALLENGE     | 4      | 3     |
| EXPOSED       | 5      | —     |

Knock-on: `SOURCE_INTERMEDIATE = CHALLENGE + 1` (was `EXPOSED + 1`); `Source::decode`'s `PREPROCESSED..=EXPOSED` range becomes `PREPROCESSED..=CHALLENGE`. `SOURCE_*` tag values shift down accordingly. There is no on-disk codec format to preserve — these tags are computed fresh per proving run.

**`crates/cuda-backend/cuda/include/codec.cuh`** — `ENTRY_PERMUTATION` and `ENTRY_EXPOSED` removed from the `EntryType` enum. The kernels that consume this enum (`batch_mle.cu`, `mle.cu`, `gkr_input.cu`, `dag_entry.cuh`) reference variants by **name**, not numeric value, so they pick up the renumbering transparently. Confirmed by grep: no kernel mentioned `ENTRY_PERMUTATION` or `ENTRY_EXPOSED`.


zlangley (approved):

crates/stark-backend/src/air_builders/symbolic/mod.rs
```diff
@@ -146,24 +125,15 @@ impl<F: Field> SymbolicRapBuilder<F> {
         if width.common_main != 0 {
             partitioned_main.push(gen_main_trace(width.cached_mains.len(), width.common_main));
         }
-        let after_challenge = Self::new_after_challenge(&width.after_challenge);
 
         let public_values = (0..num_public_values)
             .map(move |index| SymbolicVariable::new(Entry::Public, index))
             .collect();
 
-        let challenges = Self::new_challenges(num_challenges_to_sample);
```
jonathanpwang, line 155: are you sure it's ok to remove this?

stephenh-axiom-xyz: Yeah, `after_challenge`, `challenges`, and `exposed_values_after_challenge` were only used in deprecated stuff/the `PermutationAir` implementation. In `TraceWidth`, we always assert `after_challenge` is empty.

PR #338, opened 2026-05-07
feat(bench): synthetic benchmark suite for cuda-backend iteration

jonathanpwang:
## Summary

Adds a new `benchmarks/synthetic` crate (`openvm-benchmark-synthetic`) with two binaries for fast champ-vs-candidate iteration on `openvm-cuda-backend` changes, and folds the previous `benchmarks/proving` runner into it as a second binary. The previous standalone `benchmarks/proving` crate is removed.

## Motivation

Iterating on cuda-backend changes against a real reth `prove-stark` workload is slow: each candidate requires rebuilding the full prover stack and re-running the original guest, and the runtime is dominated by setup outside the kernels we actually want to measure. We want a harness that:

- Replays the *kernel-cost-faithful* shape of a real workload (per-AIR widths, heights, constraint/interaction counts) without re-running the guest.
- Is deterministic for a given `(sample_frac, seed, max_log_height)` triple so champ vs. candidate are directly comparable.
- Exposes a tiered cost/coverage knob (screening / fail-fast / full) so most iterations finish in ~60 s and only promising candidates pay the full ~11 min run.
- Has a measured jitter floor so we know when a delta is real.

The trick is the **kill column**: column 0 of every synthetic AIR is forced to zero, every constraint multiplies by it, and every interaction count uses it. Traces are all-zero, so the AIR is trivially valid but the prover does the same per-cell work it would do on a real trace. End-to-end proof validity is *not* exercised by `synthetic_runner` — we rely on the cuda-backend test suite for correctness, and on `uniform_runner` (which does verify) for a sanity check on the synthesizer itself.

## What's in here

| Path | What it is |
|---|---|
| `benchmarks/synthetic/Cargo.toml` | New crate `openvm-benchmark-synthetic` with a `cuda` feature. |
| `benchmarks/synthetic/README.md` | Full usage / workflow / scorecard docs. |
| `benchmarks/synthetic/reth-block-23992138-profile.jsonl` | Canonical baseline profile: 209 segment proofs from a real reth `prove-stark` run on block 23992138 (schema v2, 6,382 per-AIR records). |
| `benchmarks/synthetic/src/synthetic_air.rs` | `SyntheticAir` / `SyntheticShape` — the kernel-cost-faithful replay AIR. |
| `benchmarks/synthetic/src/segment_profile.rs` | Consumer-only deserialization types (`SegmentProfile`, `AirShapeRecord`); no probe/sink machinery. |
| `benchmarks/synthetic/src/bin/synthetic_runner.rs` | Profile-replay runner (GPU-only; gated on `required-features = [\"cuda\"]`). |
| `benchmarks/synthetic/src/bin/uniform_runner.rs` | Uniform-shape sweep runner — moved from `benchmarks/proving/src/main.rs`. CPU by default; switches to GPU under `--features cuda`. |
| `Cargo.toml` | Workspace members: `benchmarks/proving` removed, `benchmarks/synthetic` added. |

### `synthetic_runner` — profile replay (champ-vs-candidate)

Reads a captured `SegmentProfile` JSONL (one segment per line, each containing per-AIR shape records), seeded-samples `ceil(total * sample_frac)` segments, and for each one rebuilds a `Vec<AirRef<SC>>` of `SyntheticAir`s from the per-AIR shapes, generates all-zero traces, and runs `engine.keygen + engine.prove` on the GPU. Writes a scorecard JSON with per-segment `keygen_ms` / `prove_ms` and a top-level `total_prove_ms` figure of merit.

Tiered modes (against the bundled reth-block-23992138 profile):

| `--sample-frac` | Use case | Segments | GPU prove (ref) | Wall |
|---|---|---|---|---|
| `0.1` | screening | 21/209 | ~3.9 s | ~60 s |
| `0.5` | fail-fast | 105/209 | ~21.6 s | ~333 s |
| `1.0` | full sweep | 209/209 | ~43.6 s | ~666 s |

Wall ≫ prove because per-segment H2D transport of the all-zero traces (up to 150 M cells) dominates outside the timed `prove()` call. Comparison metric is `total_prove_ms`.

Reference jitter floor at `(0.1, 42)`: `mean=3895 ms  stddev=16 ms  CoV=0.40 %`. A real wall delta > ~0.5 % is detectable above noise.

### `uniform_runner` — uniform-shape sweep

Builds N identical synthetic AIRs sized via CLI flags (`--num-airs`, `--cols-per-air`, `--constraints-per-col`, `--interactions-per-col`, `--log-rows-per-air`, `--log-stacked-height`) and runs one keygen + prove. Useful for understanding how cost moves with a single parameter, and **verifies the proof at the end** — so it surfaces synthesizer correctness regressions that `synthetic_runner`'s profile-replay path skips.

This is the binary that used to live in `benchmarks/proving`. Behavior is unchanged; it just moved into the synthetic crate.

## Notes

- Plonky3 stays pinned to `=0.4.1`; nothing in the workspace toolchain changed.
- `synthetic_runner` does not create its `--out` parent directory — see README quick start.
- `synthetic_runner`'s `cuda` feature unconditionally pulls `openvm-cuda-backend`, hence `required-features = [\"cuda\"]`.

## Test plan

- [ ] `cargo check -p openvm-benchmark-synthetic` (CPU build)
- [ ] `cargo check -p openvm-benchmark-synthetic --features cuda` on a CUDA host
- [ ] `cargo clippy -p openvm-benchmark-synthetic --all-targets --tests -- -D warnings`
- [ ] `cargo +nightly fmt -- --check`
- [ ] On a CUDA host: build and run `synthetic_runner --sample-frac 0.1 --seed 42 --max-log-height 22` against the bundled profile; confirm scorecard `skipped_segments: 0` and `total_prove_ms` within ~0.5 % of the reference (~3895 ms).
- [ ] Run `uniform_runner` CPU and (if available) GPU at the README defaults; confirm proof verifies.
- [ ] Workspace `cargo check -p openvm-stark-backend` and `-p openvm-stark-sdk` to confirm the workspace-members edit didn't break the default build.


jonathanpwang:
Addressed review in 6173003c:

**Bug #1 — width-1 AIRs lose interaction message** ✅
Widened `SyntheticAir` width to `max(common_main, max(interaction_message_lens) + 1, 1)` in `synthetic_air.rs:from_shape`. ProgramAir-shaped records (width=1, captured 9-field msg) now widen to width=10, preserving the per-interaction kernel work. Added `width_widens_to_fit_captured_message_len` test that mirrors the ProgramAir case.

**Bug #2 — redundant `from_shape` construction** ✅
`synthetic_runner` now builds `Vec<Arc<SyntheticAir>>` once, reuses each `Arc` for both the `AirRef` keygen wrapping and the per-AIR trace generation.

**Robustness #3 — manual arg parsing** ✅
Switched `synthetic_runner` to `clap::Parser` derive (parity with `uniform_runner`), so missing values now produce real diagnostics.

**Robustness #4 — `sample_frac` validation** ✅
`eyre::ensure!(sample_frac.is_finite() && sample_frac > 0.0 && sample_frac <= 1.0)` at the top of `main`.

**Robustness #5 — `&PathBuf` → `&Path`** ✅
`read_profile` now takes `&Path`.

**Style #10/#11** ✅
Dropped the intermediate `let air: AirRef<_> = ...; air` binding and the unnecessary parens around `1u64`.

**Bonus fix** — the cuda build of `synthetic_runner` was already broken on this branch: upstream `TraceWidth` lost its `after_challenge` field in commit 610b4062, but the bundled profile JSONL still carries it. Introduced a local `ProfileTraceWidth` in `segment_profile.rs` that mirrors the captured wire format independent of upstream — old captures keep deserializing across upstream schema changes.

**Verification**
- `cargo check -p openvm-benchmark-synthetic` ✅
- `cargo check -p openvm-benchmark-synthetic --features cuda --bin synthetic_runner` ✅
- `cargo clippy -p openvm-benchmark-synthetic --features cuda --all-targets --tests -- -D warnings` ✅
- `cargo +nightly fmt -p openvm-benchmark-synthetic -- --check` ✅
- `cargo nextest run -p openvm-benchmark-synthetic` — 3/3 passing (incl. new width-1 test and the keygen→prove→verify round trip)

Non-blocking design observations (#6 binary data file, #7 keygen+PK-transport timing, #8 unused `ShapeAtlas`, #9 SDK feature note) left as-is — the `ShapeAtlas` type is forward-looking for an atlas-based replay variant; happy to remove if you'd prefer.

benchmarks/synthetic/src/synthetic_air.rs
```diff
@@ -0,0 +1,342 @@
...
+            width,
+            num_constraints: s.num_constraints,
+            max_constraint_degree: s.max_constraint_degree.max(1),
+            interactions,
+        }
+    }
+
+    pub fn width(&self) -> usize {
+        self.width
+    }
```
stephenh-axiom-xyz, line 132: nit: seems like this is a dupe function with the `BaseAir` impl below?

stephenh-axiom-xyz (approved):
LGTM, although would like to see that the synthetic benchmark time scales proportionately with observed reth benchmark times for different blocks

benchmarks/synthetic/src/synthetic_air.rs
jpw-axiom: @claude fix this

jonathanpwang: BaseAir trait needs field which is annoying

PR #339, opened 2026-05-08
feat(cuda-profiler): NVBit-based per-CTA profiler with PNG render pipeline

jonathanpwang:
## Summary

A from-scratch CUDA profiler that captures per-CTA placement (which CTA of which kernel ran on which SM, when) for **every kernel a CUDA program launches** — no source changes, no rebuild of the prover. Built around NVBit for SASS-level instrumentation (CC 7.0–12.1), libzstd for log compression, and a streaming C aggregator + Python+matplotlib renderer for offline reduction.

The two commits map to the work directly:

- `05145914` — initial source-level path (per-kernel `SHADOW_KERNEL_BEGIN/END/PARAM` macros, host ring/drain, CUPTI sidecar). Wire format and Perfetto exporter were established here.
- `c751d741` — replaces the source-level path with the NVBit tool + offline reduction pipeline. The wire format (SHDWPROF) is unchanged so the Rust `cuda-profiler-assemble` Perfetto exporter from the first commit keeps working.

## Pipeline

```
nvbit-tool.so   ─►  shadow_profile.bin (zstd-framed)  ─►  aggregate  ─►  agg.bin (~MB)  ─►  render_overview.py  ─►  trace_overview.png
(LD_PRELOAD)                                              (C, libzstd)                       (matplotlib)
```

Plus the Rust `cuda-profiler-assemble` binary that turns the same log into a Perfetto Chrome trace JSON for interactive drilldown.

## What's in `crates/cuda-profiler/nvbit-tool/`

| File | What it is |
|--|--|
| `tool.cu` | host: env-var parsing, pinned-mapped ring + drain thread, zstd-framed log writer, NVBit instrumentation hook |
| `inject_funcs.cu` | device: lane-0-only BEGIN/END probes with `atomicExch` dedup and per-CTA decimation gate |
| `aggregate.c` | streaming SHDWPROF reader (auto raw vs zstd) → compact per-(SM, time-bucket) and per-(kernel, time-bucket) grids |
| `render_overview.py` | matplotlib renderer for the two-panel PNG (kernel-composition stack + per-SM heatmap) |
| `Makefile` | builds `nvbit-tool.so` + `aggregate` |
| `README.md` | full quick-start, env-var reference, caveats |

## Log-size optimizations

Two knobs let the user trade fidelity for log size:

- **`SHADOW_PROFILER_SAMPLE_N`** — record 1 of every N CTAs from kernels with grid ≥ N (small kernels keep every CTA so they aren't lost). Power-of-two; default 1.
- **zstd-framed log output** — the on-disk file is a raw zstd frame (magic `28 b5 2f fd`); the SHDWPROF header lives inside the compressed stream. The C aggregator auto-detects raw vs zstd by peeking 4 bytes.

Together (N=8 + zstd) shrink a 25-min Reth `prove-app` log from ~58 GB to ~2-3 GB with the rendered PNG visually unchanged.

## Caveats

- **No CUPTI side-data** in the NVBit-only path. The log only carries `ProcessStart`, `KernelName`, and per-CTA records — kernel-launch / memcpy / NVTX activity is not captured. Run nsys side-by-side if you need it.
- **No GPU wall-clock kernel-time bar** in the PNG. Summing per-CTA durations over-counts parallelism; that metric was deliberately removed in favor of per-bucket-normalized composition. Use nsys / CUPTI for true wall-clock per-kernel time.
- **Rust `cuda-profiler-assemble`** still reads raw SHDWPROF only; pipe through `zstd -d` if you produce the Perfetto trace from a zstd-framed log.

## Test plan

- [ ] Local sanity: \`make ARCH=sm_90\` builds \`nvbit-tool.so\` + \`aggregate\` cleanly
- [ ] \`SHADOW_PROFILER=1 SHADOW_PROFILER_SAMPLE_N=8 LD_PRELOAD=./nvbit-tool.so ./your-cuda-binary\` produces a zstd-framed log
- [ ] \`./aggregate log.bin agg.bin\` succeeds on both raw and zstd inputs
- [ ] \`python3 render_overview.py agg.bin out.png\` produces the two-panel PNG
- [ ] cuda-backend builds without the (now-removed) \`profiler\` Cargo feature
- [ ] On a real workload (Reth prove-app block 23992138 on H100 PCIe): captured 1.46 B per-CTA records across 151 kernel families, all 114 SMs covered, 0 drops; PNG renders cleanly

closes INT-7722


jonathanpwang:
consider also: `nsys profile --gpu-metrics-devices=all`

crates/cuda-profiler/nvbit-tool/inject_funcs.cu
```diff
@@ -0,0 +1,159 @@
...
+        t_start = t_end > 0ULL ? t_end - 1ULL : 0ULL;
+    }
+    unsigned long long slot = atomicAdd(ctx->head, 1ULL);
+    CtaRecord *r = &ctx->ring[slot & static_cast<unsigned long long>(ctx->mask)];
+    r->kernel_id = kernel_id;
+    r->smid = smid;
+    r->block_linear = static_cast<uint32_t>(block_linear);
+    r->t_start = t_start;
+    r->t_end = t_end;
+    __threadfence_system();
```
Aalanli, line 155: Why is there a `__threadfence_system()`?

crates/cuda-profiler/nvbit-tool/inject_funcs.cu
```diff
@@ -0,0 +1,159 @@
...
+        return; // already emitted, or begin probe never ran for this CTA
+    }
+    uint64_t t_start = static_cast<uint64_t>(t_start_ull);
+    if (t_start > t_end) {
+        // Scratch slot was clobbered by a hash collision with a *later*
+        // CTA's begin probe. Clamp to a 1-ns span — the placement is still
+        // correct, only the duration is unreliable.
+        t_start = t_end > 0ULL ? t_end - 1ULL : 0ULL;
+    }
+    unsigned long long slot = atomicAdd(ctx->head, 1ULL);
```
Aalanli, line 148: Why are we using atomicAdd to find a slot instead of reserving per-cta space and incrementing a local counter?

crates/cuda-profiler/nvbit-tool/inject_funcs.cu
Aalanli: There should be an implicit threadfence at the kernel end.

PR #340, opened 2026-05-10
perf(cuda): reduce prover copy and batching overhead

jonathanpwang:
## Summary

This PR contains CUDA prover micro-optimizations on top of the latest `develop-v2` GKR input batching changes:

- Batch contiguous stacked trace blit copies to reduce the number of D2D memcpy launches.
- Batch stacked reduction sumcheck windows by using per-window accumulator slots and doing one D2H accumulator readback after queued kernels.
- Reduce D2H overhead for stacked openings by copying all `q_evals` into one host buffer before splitting by original lengths.
- Reuse the length-1 `EqEvalLayers` seed layer across equality trees to avoid repeated tiny device allocations/copies.
- Preserve larger batch-MLE memory budgets with 5GiB/6GiB floors so the new partial-batched GKR input path does not regress launch batching when `save_memory` changes the input-sized budget.
- Use direct stream synchronization for D2H copies.

The final optimization commit notes retained benchmark improvement on real reth same-21 `stark_prove_excluding_trace`: baseline 6241 ms, final reruns 5613 ms and 5576 ms (>10% improvement).

## Validation

Reviewed against local `develop-v2` at `c68e3923`.

- `cargo check -p openvm-cuda-backend`
- `cargo nextest run -p openvm-cuda-backend --test-threads=4` — 104 passed, 1 skipped

### DIFF
--- a/crates/cuda-backend/src/logup_zerocheck/mod.rs
+++ b/crates/cuda-backend/src/logup_zerocheck/mod.rs
@@ -93,10 +93,16 @@ use round0::evaluate_round0_constraints_gpu;
 /// This ratio can be tuned. Currently it is set to prefer the monomial kernel except for the
 /// Poseidon2Air where the DAG node size is much smaller than number of monomials.
 const DAG_FALLBACK_MONOMIAL_RATIO: usize = 2;
-// Batch MLE launch overhead dominates when the budget follows only the GKR input size.
-// Keep at least the historical non-save-memory batch budget while still allowing larger inputs.
-const BATCH_MLE_MEMORY_LIMIT_FLOOR: usize = 5 << 30; // 5GiB
-const BATCH_MLE_FOLD_MEMORY_LIMIT_FLOOR: usize = 6 << 30; // 6GiB
+// Batch MLE launch overhead dominates in the non-save-memory path, so keep the historical
+// fixed batch budget there.
+const BATCH_MLE_DEFAULT_MEMORY_LIMIT: usize = 6 << 30; // 6GiB
+
+#[inline]
+fn fractional_gkr_peak_memory_bytes(input_len: usize, peak_work_buffer_bytes: usize) -> usize {
+    input_len
+        .saturating_mul(std::mem::size_of::<Frac<EF>>())
+        .saturating_add(peak_work_buffer_bytes)
+}
 
 #[inline]
 pub(crate) fn air_width_for_mat(need_rot: bool, mat_width: usize) -> u32 {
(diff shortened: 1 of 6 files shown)


crates/cuda-backend/src/logup_zerocheck/mod.rs
```diff
@@ -202,14 +206,15 @@ where
...
+        if prover.memory_limit_bytes > BATCH_MLE_MEMORY_LIMIT_FLOOR {
             tracing::warn!(
                 "prover.memory_limit_bytes {} already exceeds 5GiB",
                 prover.memory_limit_bytes
             );
         }
-        prover.memory_limit_bytes = DEFAULT_MEMORY_LIMIT;
+        prover.memory_limit_bytes = BATCH_MLE_MEMORY_LIMIT_FLOOR;
+    } else {
+        prover.memory_limit_bytes = prover.memory_limit_bytes.max(BATCH_MLE_MEMORY_LIMIT_FLOOR);
```
gaxiom: Does this break the purpose of the flag 'save_memory=true'?

crates/cuda-backend/src/poly.rs
```diff
@@ -379,8 +379,29 @@ impl EqEvalSegments<EF> {
 /// dropping layers.
 #[derive(Getters)]
 pub struct EqEvalLayers<F> {
-    /// Index 0 should never to read, but it will be initialized to zero.
-    pub layers: Vec<DeviceBuffer<F>>,
+    /// Layer 0 stores the length-1 equality value and may be shared across trees.
+    pub layers: Vec<EqEvalLayer<F>>,
+}
+
+pub enum EqEvalLayer<F> {
```
gaxiom: Why not just to use `Arc<DeviceBuffer>` always?

gaxiom (approved):
LGTM

crates/cuda-backend/src/logup_zerocheck/mod.rs
jonathanpwang: fixed and increased the gkr limit using the `peak_work_buffer_size` helper function

crates/cuda-backend/src/stacked_pcs.rs
```diff
@@ -154,27 +154,45 @@ pub(crate) fn stack_traces_into_expanded(
...
+            let mut end = idx + 1;
+            while end < layout.sorted_cols.len() {
+                let (next_mat_idx, next_j, next_s) = &layout.sorted_cols[end];
+                if *next_mat_idx != *mat_idx || next_s.log_height() != s.log_height() {
+                    break;
+                }
+                let expected_j = *j + (end - idx);
+                let next_len = next_s.len(l_skip);
+                let next_start = next_s.col_idx * padded_height + next_s.row_idx;
+                if *next_j != expected_j || next_len != s_len || next_start != start + copy_len {
```
jonathanpwang, line 176: @claude shouldn't this be unreachable?

PR #343, opened 2026-05-13
perf(cuda): make bn254_grind kernel faster

Aalanli:
# main reason for perf 

Due to excessive use of `__forceinline__`, many instructions generated in 128 bit arithmetic and branching nature of the grind kernel, instructions gets duplicated significantly, poor perf comes from instruction cache miss for the extremely large number of instructions in this kernel. 

Notably nvcc doesn't do a good job of knowing when or when not to inline by default (removing `__forceinline__`), likely because this is very out of distribution, as a small number of functions generate large blobs of ptx that gets repeated many times. So I have to manually insert noinline pragmas. 

There's a tradeoff between inlining, register usage and instruction size. Inlining reduces register pressure for instruction bloat. After the noinline optimization, it is my observation that the large amount of local thread state for this hash function creates a lot of register spills, so the kernel is actually memory bound, after taking care of the instruction cache miss.

NCU reveals (for the kernel in this PR)
image

image

The local register spills take up L1/L2 bandwidth. Further performance optimization should aim to reduce register spilling with shared memory.

# perf (on synthetic)

new:
```
loaded 209 segments from benchmarks/synthetic/reth-block-23992138-profile.jsonl
sampled 11/209 segments (frac=0.050, seed=42, max_log_h=22)
[  1/ 11] seg   2 | 20 airs |   440315986 cells | keygen    13 ms | prove   1401 ms
[  2/ 11] seg  34 | 23 airs |   482928660 cells | keygen    81 ms | prove   1054 ms
[  3/ 11] seg  62 | 24 airs |   401136404 cells | keygen    84 ms | prove   1293 ms
[  4/ 11] seg  92 | 34 airs |   429012500 cells | keygen   126 ms | prove   1575 ms
[  5/ 11] seg  98 | 34 airs |   398776112 cells | keygen   128 ms | prove    952 ms
[  6/ 11] seg 121 | 19 airs |   469386260 cells | keygen    11 ms | prove   1771 ms
[  7/ 11] seg 139 | 23 airs |   497466388 cells | keygen    82 ms | prove   1041 ms
[  8/ 11] seg 140 | 23 airs |   437625876 cells | keygen    85 ms | prove   1053 ms
[  9/ 11] seg 150 | 42 airs |   143752444 cells | keygen    51 ms | prove    728 ms
[ 10/ 11] seg 166 | 42 airs |   155843836 cells | keygen    51 ms | prove    687 ms
[ 11/ 11] seg 196 | 42 airs |    78153852 cells | keygen    50 ms | prove    558 ms
```

old:
```
loaded 209 segments from benchmarks/synthetic/reth-block-23992138-profile.jsonl
sampled 11/209 segments (frac=0.050, seed=42, max_log_h=22)
[  1/ 11] seg   2 | 20 airs |   440315986 cells | keygen    12 ms | prove  27004 ms
[  2/ 11] seg  34 | 23 airs |   482928660 cells | keygen    81 ms | prove  26631 ms
[  3/ 11] seg  62 | 24 airs |   401136404 cells | keygen    83 ms | prove  27602 ms
[  4/ 11] seg  92 | 34 airs |   429012500 cells | keygen   126 ms | prove  28353 ms
[  5/ 11] seg  98 | 34 airs |   398776112 cells | keygen   127 ms | prove  26338 ms
[  6/ 11] seg 121 | 19 airs |   469386260 cells | keygen    12 ms | prove  27115 ms
[  7/ 11] seg 139 | 23 airs |   497466388 cells | keygen    84 ms | prove  26633 ms
[  8/ 11] seg 140 | 23 airs |   437625876 cells | keygen    84 ms | prove  26615 ms
[  9/ 11] seg 150 | 42 airs |   143752444 cells | keygen    51 ms | prove  26087 ms
[ 10/ 11] seg 166 | 42 airs |   155843836 cells | keygen    51 ms | prove  26268 ms
[ 11/ 11] seg 196 | 42 airs |    78153852 cells | keygen    50 ms | prove  25988 ms
```

So between 15-45x improvement.

**note**
I created inline and noinline versions of the bn254 header file because for the grind kernel we need noinline due to large instruction blowup otherwise, while for the `bn254_compressing_row_hashes_kernel`, we need inlining because of register pressure. 

Resolves INT-7806


crates/cuda-backend/src/bn254_sponge.rs
```diff
@@ -260,15 +262,15 @@ mod tests {
...
+            FiatShamirTranscript::<BabyBearBn254Poseidon2Config>::observe(&mut cpu, witness);
+            let witness_bits =
+                FiatShamirTranscript::<BabyBearBn254Poseidon2Config>::sample_bits(&mut cpu, bits);
             // Verify the CUDA-found witness passes check_witness on the CPU transcript.
             assert!(
-                FiatShamirTranscript::<BabyBearBn254Poseidon2Config>::check_witness(
-                    &mut cpu, bits, witness
-                ),
-                "CUDA witness {witness:?} invalid on CPU (observed {num_observed} values)"
+                witness_bits == 0,
```
jonathanpwang: nit (but I guess clippy didn't care): clippy often tells you to use `assert_eq`

jonathanpwang:
I'll let @gaxiom review this one.

Also FYI, claude posts a new comment each time, so you have to manually hide the old ones... (we just haven't gotten around to patching the workflow)

gaxiom:
### Code review

Found 5 issues:

1. Stale file-level doc comment in `poseidon2_bn254.cuh` — after this PR, `Bn254Fr` and the field constants live in `poseidon2_bn254_common.cuh`, but the header still says "This header provides: Bn254Fr type..." and "The Poseidon2 permutation function is declared here but defined in bn254_poseidon2.cu" (no permutation is declared in this header).

https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/cuda/include/poseidon2_bn254.cuh#L1-L18

2. Stale doc comment copied verbatim into the new `poseidon2_bn254_noinline.cuh`. Same wrong claim about `Bn254Fr` and the permutation; this header only contains field arithmetic in `namespace bn254_noinline`.

https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/cuda/include/poseidon2_bn254_noinline.cuh#L1-L20

3. `bn254_grind` benchmark double-syncs and mismeasures. The bench calls `dev.sync_h2d(&ctx)` outside the timer, but `grind_gpu` itself calls `self.sync_h2d(device_ctx)?` (bn254_sponge.rs:209) inside the timed region — so the explicit pre-sync is dead and `t0.elapsed()` includes an H2D copy that the PR description's "27s → 1.4s" numbers don't disclose.

https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/benchmarks/synthetic/src/bin/bn254_grind.rs#L15-L26

caller it duplicates: https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/src/bn254_sponge.rs#L200-L212

4. `#pragma unroll 1` was added to the partial-rounds loop of `bn254_poseidon2_permute_impl` (the `__forceinline__` template). That template is instantiated for both `bn254_poseidon2_permute<3,4,56>` (used by row hashes / sponge) and `bn254_poseidon2_permute_w2<2,3,50>`. The PR description only benchmarks the grind kernel (which goes through the new `_implv2`); the row-hash and width-2 Merkle-compress paths now also lose unrolling on their inner partial loop and aren't measured. Either move the pragma onto `_implv2` only, or include a row-hash NCU profile in the PR.

https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/cuda/src/bn254_poseidon2.cu#L80-L100

5. `load_shared()` cooperatively fills `__shared__ buf` from three strided loops and returns pointers into it without an internal `__syncthreads()`. The single caller adds the barrier on line 531, so today this is correct, but the function silently relies on every future caller to remember; both this and the dead `__constant__` `BN254_R_ONE` / `BN254_R_TWO` look like leftover scaffolding worth cleaning up before merge.

https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/cuda/src/bn254_poseidon2.cu#L36-L56

dead constants: https://github.com/openvm-org/stark-backend/blob/efb82065bacba2aef5e40aff9a95ccb4ce5a1a25/crates/cuda-backend/cuda/include/poseidon2_bn254_common.cuh#L37-L55

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Aalanli (commented):
Fixed minor typos. 
Comments 3, 5 not relevant.
3: PR numbers wern't derived from bn254_grind. Added pre-sync to populate device buffers.
5: put syncthreads after the caller because there's additional setup before, so this gets rid of one sync.

Aalanli (commented):
Also volatile is necessary for correctness. Loads may be cached in L1.

Aalanli:
@gaxiom , can you take a look?

crates/cuda-backend/cuda/include/poseidon2_bn254.cuh
```diff
@@ -13,66 +12,16 @@
 /// function is declared here but defined in bn254_poseidon2.cu (where the
```
gaxiom: I don't think that it's still valid. Please check

crates/cuda-backend/cuda/include/poseidon2_bn254_noinline.cuh
```diff
@@ -0,0 +1,266 @@
...
+/// This header provides:
+///   - CIOS Montgomery multiplication matching p3-bn254::helpers::monty_mul
+///   - Field add / sub / neg / double / x^5 S-box
+///   - Templated MDS layers (external and internal, any WIDTH)
+///   - bn254_from_canonical / bn254_to_canonical conversions
+///   - bn254_pack_base_2_31 / u256_mod_u32 helpers for MultiFieldTranscript
+///     packing and sampling during grinding
+///
+/// All functions are pure computation (no globals). The Poseidon2 permutation
+/// function is declared here but defined in bn254_poseidon2.cu (where the
```
gaxiom: Same here. IS this comment correct?

benchmarks/synthetic/src/bin/bn254_grind.rs
```diff
@@ -0,0 +1,27 @@
...
+    /// number of bits in PoW to zero
+    #[arg(long, default_value_t = 18)]
+    bits: usize,
+}
+
+fn main() {
+    let Args { bits } = Args::parse();
+    let ctx = GpuDeviceCtx::for_device(get_device().unwrap() as u32).unwrap();
+    let mut dev = MultiFieldTranscriptGpu::default();
+    dev.sync_h2d(&ctx).unwrap();
```
gaxiom, line 19: Is it on purpose outside of the timer?

crates/cuda-backend/cuda/src/bn254_poseidon2.cu
```diff
@@ -29,6 +30,29 @@ __device__ __constant__ Bn254Fr g_partial_rc[56];
...
+    }
+
+    for (int i = threadIdx.x; i < 56 * 4; i += blockDim.x) {
+        buf[i + 12 * 4] = ((uint64_t *)g_partial_rc)[i];
+    }
+
+    for (int i = threadIdx.x; i < 12 * 4; i += blockDim.x) {
+        buf[i + 12 * 4 + 56 * 4] = ((uint64_t *)g_terminal_rc)[i];
+    }
+    auto ptr = (Bn254Fr *)buf;
```
gaxiom, line 53: If you fill shared buffer, don't you need to `__syncthreads()` after? Better to do that in this method so everyone that uses it can be sure that data is there

crates/cuda-backend/src/bn254_sponge.rs
```diff
@@ -246,6 +247,7 @@ mod tests {
     #[test]
     fn test_grind_gpu_witness_valid_on_cpu() {
         let bits = 8;
+        let mut ctx = GpuDeviceCtx::for_device(get_device().unwrap() as u32).unwrap();
```
gaxiom: why `mut`?

gaxiom (approved):
LGTM with minor fixes

benchmarks/synthetic/src/bin/bn254_grind.rs
Aalanli: Yes, to move device buffer out

crates/cuda-backend/cuda/src/bn254_poseidon2.cu
Aalanli: I think since it's local to this file it should be fine. I also want to batch multiple shared memory loads so it makes sense to put __syncthreads() in the caller.

crates/cuda-backend/src/bn254_sponge.rs
Aalanli: I guess linter didn't catch, changing.

PR #347, opened 2026-05-19
feat: optimize instruction sequence of bn254 hash functions

Aalanli:
# Summary

Because the compressing row hashes kernel is compute bound, this PR aims to optimize the SASS instruction sequence for the core bn254 hash functions (by using inline PTX). 

The optimizations on a high level are as follows:
1. use GPU friendly 32 bit arithmetic for the extended precision math; there are SASS intrinsics which support this use case through carry chains for example.
2. reduce register spilling by selectively caching in shared memory.

Overall the gains are from reduced number of instructions executed and memory accesses, although the former gives greater impact. 

I split the original poseidon2 source into 3 files, 1 for the permutation constants (`bn254_constants.cu`), 1 for the grind kernel (`bn254_poseidon2_grind.cu`) and 1 for the row hash functions (`bn254_poseidon2_row_hash.cu`), because the grind kernel uses it's own noinline header file (`poseidon2_bn254_noinline.cuh`) and is memory bound, while the compressing row hash functions are compute bound and uses it's 32 bit header (`poseidon2_bn254_b32.cuh`).

# Overall perf on synthetic benchmark

```
  [  1/ 21] seg   2 | 20 airs |   440315986 cells | prove1   1461 ms | prove2   1242 ms | 1.18x
  [  2/ 21] seg   5 | 19 airs |   346005524 cells | prove1   1158 ms | prove2    996 ms | 1.16x
  [  3/ 21] seg  34 | 23 airs |   482928660 cells | prove1    991 ms | prove2    810 ms | 1.22x
  [  4/ 21] seg  43 | 34 airs |   616338836 cells | prove1   2538 ms | prove2   1989 ms | 1.28x
  [  5/ 21] seg  62 | 24 airs |   401136404 cells | prove1   1538 ms | prove2   1444 ms | 1.07x
  [  6/ 21] seg  83 | 24 airs |   394992404 cells | prove1    796 ms | prove2    703 ms | 1.13x
  [  7/ 21] seg  88 | 24 airs |   496912924 cells | prove1   2356 ms | prove2   1468 ms | 1.60x
  [  8/ 21] seg  92 | 34 airs |   429012500 cells | prove1    886 ms | prove2    884 ms | 1.00x
  [  9/ 21] seg  98 | 34 airs |   398776112 cells | prove1    888 ms | prove2    764 ms | 1.16x
  [ 10/ 21] seg 121 | 19 airs |   469386260 cells | prove1    870 ms | prove2    731 ms | 1.19x
  [ 11/ 21] seg 130 | 23 airs |   282002548 cells | prove1    921 ms | prove2    711 ms | 1.30x
  [ 12/ 21] seg 139 | 23 airs |   497466388 cells | prove1   1429 ms | prove2   1001 ms | 1.43x
  [ 13/ 21] seg 140 | 23 airs |   437625876 cells | prove1   1175 ms | prove2    902 ms | 1.30x
  [ 14/ 21] seg 150 | 42 airs |   143752444 cells | prove1    767 ms | prove2    665 ms | 1.15x
  [ 15/ 21] seg 166 | 42 airs |   155843836 cells | prove1    681 ms | prove2    581 ms | 1.17x
  [ 16/ 21] seg 175 | 42 airs |   150830332 cells | prove1    607 ms | prove2    515 ms | 1.18x
  [ 17/ 21] seg 185 | 42 airs |   155991292 cells | prove1    601 ms | prove2    514 ms | 1.17x
  [ 18/ 21] seg 195 | 42 airs |    78153852 cells | prove1    865 ms | prove2    463 ms | 1.87x
  [ 19/ 21] seg 196 | 42 airs |    78153852 cells | prove1    574 ms | prove2    470 ms | 1.22x
  [ 20/ 21] seg 202 | 42 airs |    55320188 cells | prove1    417 ms | prove2    360 ms | 1.16x
  [ 21/ 21] seg 206 | 42 airs |    28790946 cells | prove1    821 ms | prove2    452 ms | 1.82x
```

# Perf analysis
(green is old, blue is new)

As expected, decreased compute because less instructions and decreased memory because no register spilling.
image

Completely remove local memory requests:
image
(one concern is the ICC miss, which due to the high instruction count causes instruction cache misses, as we see on multiple levels. My experiments with `noinline` yields worse performance because `noinline` causes register spills)

After this optimization, this is the synthetic benchmark profile:
image
(The warp occupancy is low because the hash function uses a lot of registers per warp.)

Unfortunately there isn't a lot you could do to the compressing_row_hashes kernel because it processes a lot of data   in the trace matrix (especially the first big one). NVCC is also very fragile when it comes to register allocation, doing minor changes can completely break it's heuristics and cause register spills which is bad for perf. I think a better approach to optimize these types of kernels is to generate the PTX ourselves and leverage a super optimizer to get the best instruction sequence, but that will likely not yield significant gains.

Resolves INT-7845


crates/cuda-backend/cuda/src/bn254_poseidon2_row_hash.cu
```diff
@@ -0,0 +1,486 @@
...
+
+extern "C" int _bn254_poseidon2_compressing_row_hashes(
+    bn254_digest_t *out,
+    const Fp *matrix,
+    size_t width,
+    size_t query_stride,
+    size_t log_rows_per_query,
+    cudaStream_t stream
+) {
+    if (log_rows_per_query > 4) {
```
gaxiom: Wait. Are we not allowed values up to 10 anymore? Should reflect it in merkle_tree.rs

crates/cuda-backend/cuda/include/poseidon2_bn254_b32.cuh
```diff
@@ -0,0 +1,411 @@
...
+
+// ---------------------------------------------------------------------------
+// Montgomery helpers (32-bit-limb, inline PTX, rhs as u32)
+//
+// rhs is a single 32-bit limb, so:
+//   - mul_small computes lhs * rhs  (256 × 32 → 288 bits = 9 × u32 limbs)
+//   - mul_small_and_acc computes lhs * rhs + add  (still ≤ 2^288, so 9 limbs)
+//
+// Both functions return the lowest 32-bit limb (r[0]) and write the upper
+// 8 limbs (r[1..8]) into `high`. Each iteration uses one `mad.wide.u32 t,
```
gaxiom: You have implemented it not in a way how it is described in the comment, isn't it?

crates/cuda-backend/cuda/src/bn254_poseidon2_row_hash.cu
```diff
@@ -0,0 +1,486 @@
...
+// ---------------------------------------------------------------------------
+// BN254 Merkle digest: a single Bn254Fr element (32 bytes)
+// Matches Digest = [Bn254Scalar; 1] on the Rust side.
+// ---------------------------------------------------------------------------
+
+struct bn254_digest_t {
+    Bn254Fr elem;
+};
+
+static const int BN254_BABY_BEAR_RATE = 16;
```
gaxiom: Better to have it in bn254_common then to redefine in multiple places

crates/cuda-backend/cuda/src/bn254_poseidon2_row_hash.cu
```diff
@@ -0,0 +1,486 @@
...
+            state[i] = bn254_b32::bn254_add(state[i], terminal_rc[r * WIDTH + i]);
+            state[i] = bn254_b32::bn254_sbox(state[i]);
+        }
+        bn254_b32::bn254_mds_external<WIDTH>(state);
+    }
+}
+
+static __device__ void bn254_poseidon2_permute_b32(Bn254Fr32 state[3]) {
+    bn254_poseidon2_permute_impl_b32<3, 4, 56>(
+        state, (Bn254Fr32 *)g_initial_rc, (Bn254Fr32 *)g_partial_rc, (Bn254Fr32 *)g_terminal_rc
```
gaxiom, line 77: Just a question: are you guarantee that 4*u64 interpreted as 8*u32 will not be silently corrupted?

gaxiom (approved):
PR is solid, just please react to small findings

crates/cuda-backend/cuda/include/poseidon2_bn254_b32.cuh
Aalanli: Sorry, I forgot to change when I switched the impls.

crates/cuda-backend/cuda/src/bn254_poseidon2_row_hash.cu
Aalanli: Yes, they have the same memory layouts.

crates/cuda-backend/cuda/src/bn254_poseidon2_row_hash.cu
Aalanli: Removed template as it wasn't necessary for perf. Changed back

PR #352, opened 2026-06-08
feat: add segment memory metering config

shuklaayush:
resolves int-8128

### DIFF
--- a/crates/stark-backend/src/memory_metering.rs
+++ b/crates/stark-backend/src/memory_metering.rs
@@ -1,19 +1,19 @@
-//! Memory estimates for segmented proving.
+//! Memory estimates for proving.
 
 use std::{cmp::max, mem::size_of};
 
 use crate::{StarkProtocolConfig, SystemParams};
 
-/// Fixed per-segment interaction scratch not proportional to interaction cells.
-const DEFAULT_INTERACTION_MEMORY_OVERHEAD: usize = 2 << 20;
+/// Fixed interaction scratch not proportional to interaction cells.
+pub const INTERACTION_MEMORY_OVERHEAD: usize = 2 << 20;
 
-/// Per-segment cell counts derived from trace heights.
+/// Cell counts for a proving memory estimate.
 ///
 /// `main_cells_*` are `Σ(padded_height * width)` in base-field cells, split by whether a trace
 /// opens next-row rotations. `interaction_cells` is the metered row-interaction slot count after
 /// power-of-two trace padding.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
-pub struct SegmentMemoryCounts {
+pub struct ProvingMemoryCounts {
     /// Main trace cells for AIRs that open next-row rotations after power-of-two padding.
     pub main_cells_with_rot: usize,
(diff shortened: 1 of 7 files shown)


crates/stark-backend/src/memory_metering.rs
```diff
@@ -0,0 +1,282 @@
+//! Memory estimates for segmented proving.
+
+use std::{cmp::max, mem::size_of};
+
+use crate::{StarkProtocolConfig, SystemParams};
+
+/// Fixed per-segment interaction scratch not proportional to interaction cells.
+const DEFAULT_INTERACTION_MEMORY_OVERHEAD: usize = 2 << 20;
+
+/// Per-segment cell counts derived from trace heights.
```
jonathanpwang: as a matter of purity, stark-backend does not know about segments, so use a more neutral name

crates/stark-backend/src/memory_metering.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Size of one base-field element in bytes.
+    pub base_field_size: usize,
+    /// Degree of the extension field over the base field.
+    pub extension_degree: usize,
+    /// `-log_2` of the rate for the initial Reed-Solomon code.
+    pub log_blowup: usize,
+    /// Whether the prover keeps the Reed-Solomon code matrix cached after `stacked_commit`.
+    pub cache_rs_code_matrix: bool,
+    /// Secondary memory contribution per main cell, per PCS opening, in base-field cells.
+    pub main_cell_secondary_weight: f64,
```
jonathanpwang: I feel like now that we are in the same crate as the actual implementation, this shouldn't be needed so much? maybe there's some issues with cuda-backend being a separate crate. I had hoped that we could tie the actual function used in cuda-backend to the memory measurement. But since we want to avoid importing cuda-backend in stark-backend, that seems impossible.

Not sure what the best solve is here, but at least we should have a unit test in cuda-backend that somehow ties the code there to this struct. And some enforcement mechanism to ensure we keep them synchronized.

jonathanpwang (commented):
I'm somewhat not happy with how the code around the memory config is not tied to the actual cuda code in any way. This is still better than having it in openvm, but can you think a bit more about how we could enforce the connection between the memory config <> cuda code better?

crates/stark-backend/src/memory_metering.rs
shuklaayush: good call, renamed to `ProvingMemory`

crates/stark-backend/src/memory_metering.rs
shuklaayush: removed the weight etc. parameters, and tried to derive the other parameters from the config/system parameters

shuklaayush: > I had hoped that we could tie the actual function used in cuda-backend to the memory measurement. But since we want to avoid importing cuda-backend in stark-backend, that seems impossible.

yeah, i don't think this is possible

jonathanpwang (approved):
I guess add to AGENTS.md or somewhere in the cuda code around the gkr section that we need to update the estimates whenever cuda code changes

PR #354, opened 2026-06-09
ci: migrate RunsOn labels for v3

shuklaayush:


.github/workflows/docs.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}
     runs-on:
-      - runs-on=${{ github.run_id }}/runner=test-gpu-nvidia
+      - runs-on=${{ github.run_id }}/image=ubuntu24-gpu-x64/family=g6.*+g5+g4dn+g6e/cpu=4+32/volume=40gb/spot=capacity-optimized/extras=s3-cache
```
jonathanpwang: actually this is a good time to just drop support for g4dn; it's older and we don't want to run into things where it gives old cuda behavior we don't care about

jonathanpwang: if you just don't specify volume, is there a default?

jonathanpwang (approved):

.github/workflows/docs.yml
shuklaayush: probably but i think it depends on the instance what the value will be? not sure

PR #359, opened 2026-06-15
feat(keygen): reject degenerate circuits

zlangley:
Require every AIR to have positive width, at least one constraint or interaction, and every interaction message to have positive length.

### DIFF
--- a/crates/stark-backend/src/keygen/mod.rs
+++ b/crates/stark-backend/src/keygen/mod.rs
@@ -10,7 +10,7 @@ use crate::{
     air_builders::symbolic::{
         get_symbolic_builder,
         symbolic_variable::{Entry, SymbolicVariable},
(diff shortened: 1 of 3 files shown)


crates/stark-backend/src/keygen/mod.rs
```diff
@@ -294,13 +313,12 @@ impl<SC: StarkProtocolConfig> AirKeygenBuilder<SC> {
         })
     }
 
-    pub fn get_symbolic_builder(&self) -> SymbolicRapBuilder<SC::F> {
```
stephenh-axiom-xyz, line 297: Does this not have any downstream usage?

stephenh-axiom-xyz (approved):
LGTM after addressing comment

crates/stark-backend/src/keygen/mod.rs
zlangley: Not for us, at least.

PR #362, opened 2026-06-18
fix: cuda concurrency issues

Aalanli:
Resolves: INT-8434

This PR fixes 2 issues, with with VPMM and one with cuda concurrency in the downstream use of the GPUEngine.

# VPMM
In the vpmm memory allocator vm_pool.rs line 672. We are defragmenting because an allocation was too big to fit in free regions. First we allocate fresh memory equal the the requested size minus the free region, then we synchronize the free regions and reuse them.

The problem I think is that the synchronization before reuse is not correct as the code calls stream.wait(&region.event) before calling vpmm apis to remap the free region address range. stream.wait makes the stream wait but is not synchronous on host, while vpmm_map is synchronous wrt. the host. So what can happen is that a kernel is not yet done executing on a free region (since it was freed asynchronously), the code calls stream.wait(...) which does not pause the execution of the host and wait for the kernel to be done, then remap_regions calls vpmm_set_access which remaps the memory address range before the kernel is actually done with it.

I think the above is undefined behavior.

# GPUEngine
What happens is that in the prover the trace gen and proving each constructs a separate GPUEngine with it's own stream. This causes two concurrency problems.

1. the work is launched in stream A in trace gen and the buffers are passed to stream B in proving, there's no explicit synchronization between them I think.
2. the trace gen buffers are allocated on stream A, the work being done on them are launched on stream B. When buffers on A are dropped using cudaFreeAsync with stream A, which does not wait for work on B to complete. This is the cause of the illegal memory access as increasing page size increases the number of allocations using cudaMallocAsync/cudaFreeAsync, which surfaces the issue on compute-sanitizer when the vpmm allocator doesn't. 

# Fix
I think a fix would be to do as pytorch does. There is a default stream that everyone launches work on, all memory allocations are allocated on the default stream, and all non default streams have to synchronize wrt. the default stream. On deallocation, if the stream that that the buffer is allocated on is not the default stream, then sync with the host because we can't be sure what work is dependent on the buffer but on a different stream.

**case 1**:
suppose bufA is allocated on stream A.
```
kernelB<<<..., defaultStream>>>(..., bufA);

drop(bufA); // kernelB could still be running on defaultStream when this is dropped. So we must make defaultStream depend on stream A and async free on the default stream. 
```
**case 2**:
suppose bufA is allocated on stream A.
```
kernelB<<<..., streamB>>>(..., bufA);
default_stream().wait_on(streamB); // this must occur as otherwise the work on streamB is independent of defaultStream and stream A, so work could still be happening as it's being dropped.
// streamA.wait_on(streamB); <- this could also work. 

drop(bufA);
```
**case 3**:
suppose buf is allocated on defaultStream.
```
kernelB<<<..., streamB>>>(..., buf);

default_stream().wait_on(streamB); // this must occur, same reasoning as above.
drop(buf);
```

The changes here protect against case 1. For cases 2,3 there's no way of protecting against race conditions in the allocator without a full device sync (sync all the streams), as any arbitrary work could be launched on a separate stream that depends on the current buffer. So it's up to the programmer to make the work dependent either on the stream the buffer is allocated on or the default stream, before any buffers the kernel launches is dropped.

### DIFF
--- a/crates/cuda-common/src/memory_manager/mod.rs
+++ b/crates/cuda-common/src/memory_manager/mod.rs
@@ -118,15 +118,18 @@ impl MemoryManager {
     unsafe fn d_free_under_lock(&mut self, ptr: *mut c_void) -> Result<StreamGuard, MemoryError> {
         let nn = NonNull::new(ptr).ok_or(MemoryError::NullPointer)?;
 
+        let default_stream = default_stream();
         if let Some(record) = self.allocated_ptrs.remove(&nn) {
-            if record.stream != default_stream() {
+            if record.stream != default_stream {
                 // we need to synchronize any memory allocated that is not on the default stream
                 // as work launched prior to drop might depend on this memory
-                record.stream.synchronize().map_err(MemoryError::Cuda)?;
+                default_stream
+                    .wait_on(&record.stream)
+                    .map_err(MemoryError::Cuda)?;
             }
             let size = record.size;
             self.current_size -= size;
-            check(unsafe { cudaFreeAsync(ptr, record.stream.as_raw()) }).map_err(|e| {
+            check(unsafe { cudaFreeAsync(ptr, default_stream.as_raw()) }).map_err(|e| {
                 tracing::error!("cudaFreeAsync failed: ptr={:p}: {:?}", ptr, e);
                 MemoryError::from(e)
             })?;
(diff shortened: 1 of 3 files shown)


Aalanli:
reth benchmark on changes made in this branch: https://github.com/axiom-crypto/openvm-eth/actions/runs/27785218280

stephenh-axiom-xyz (commented):
We are going to not merge this yet, as it has a pretty large blast radius. Let's look to merge it after the v2.0.0 release

yi-sun:
Will let @gaxiom weigh in as well, but we may want to consider these stream / memory allocation issues as a whole across the codebase.

stephenh-axiom-xyz:
Closing in preparation for v2 release, will revisit later

PR #366, opened 2026-06-26
feat: decrease root pow bits

Aalanli:
Also export more things to be public for easier testing of different params downstream in openvm.

On reth-bench new root settings lower 0.74s from 1.44s. This is because pow kernel takes a large fraction of the time.

new: https://github.com/axiom-crypto/openvm-eth/actions/runs/28200719309
orig: https://github.com/axiom-crypto/openvm-eth/actions/runs/27851015413

Resolves: INT-7945


zlangley (approved):
LGTM

PR #367, opened 2026-06-27
fix: tighten CUDA proving memory estimates

shuklaayush:
## Summary

Tightens CUDA proving-memory estimates so metered batching accounts for retained CUDA prover state and fractional-GKR scratch instead of relying on the older interaction-cell weight.

The estimate now:

- separates opened main cells from common-main stacked cells, so RS-code and retained commitment memory use the same shape as the prover
- includes CUDA-only retained common-main commitment / WHIR opening memory when building a CUDA prover config
- shares fractional-GKR work-buffer sizing from the prover-owned GKR sizing model, so CUDA batching and memory metering use the same Rust arithmetic
- keeps generic `ProvingMemoryConfig` unchanged by default; CUDA opts in through `with_cuda_backend(cache_stacked_matrix)`

The CUDA prover workflow is not changed. The CUDA-side diff only wires the existing prover cache setting into metering and delegates batching work-buffer sizing to the shared prover-owned helper.

## Estimate Shape

Main opened trace memory is always resident for the segment:

```text
main_cells = main_cells_with_rot + main_cells_without_rot
main       = main_cells * sizeof(F)
```

CUDA retained common-main state is only enabled for CUDA configs:

```text
main_stacked_cells = stacked_width * 2^log_stacked_height
cached_stacked     = main_stacked_cells * sizeof(F), when cache_stacked_matrix
common_digest      = retained Merkle digest layers for the common-main commitment
retained_common    = cached_stacked + common_digest
```

Secondary main-trace buffers follow the existing opening/interpolation weight:

```text
mat_eval_bytes    = (padded_height / 2^l_skip) * ((1 + need_rot) * width) * sizeof(EF)
secondary_weight  = (1 + constraint_degree / 2) * D_EF / 2^l_skip
main_secondary    = sizeof(F) *
                    (2 * secondary_weight * main_cells_with_rot
                     + secondary_weight * main_cells_without_rot)
```

Interaction memory now uses the fractional-GKR sizing model:

```text
gkr_input   = 2 * interaction_cells * D_EF * sizeof(F)
gkr_work    = max(fold_eval_work_buffer, precompute_m_work_buffer)
interaction = gkr_input + gkr_work + max(2 MiB, round_temp_buffer)
```

The total estimate is `main` plus the largest secondary phase:

```text
commit_peak = cached_stacked + if cache_rs_code_matrix {
    rs_code_matrix + common_digest
} else {
    max(rs_code_matrix + common_first_layer, common_digest)
}

[...]

PR #370, opened 2026-06-29
fix: align CUDA fractional GKR memory metering

shuklaayush:
## Summary

Follow-up to #367. That PR made CUDA proving memory estimates more faithful, but the fractional-GKR interaction estimate still overcharged default CUDA runs by treating two mutually exclusive work-buffer paths as if the larger path always applied:

- fold-eval work buffer: `logical_len / 4` `Frac` elements
- precompute-M work buffer: `max(logical_len >> (1 + GKR_WINDOW_SIZE), 2^GKR_WINDOW_DEFAULT_MIN_N)` `Frac` elements, plus EF auxiliary buffers

CUDA defaults to precompute-M through `SWIRL_CUDA_GKR_PRECOMPUTE_M`, so default proving should meter the precompute-M budget instead of the fold-eval budget. This PR wires the CUDA-selected fractional-GKR strategy into `ProvingMemoryConfig` and keeps the generic `openvm-stark-backend` default conservative.

This is a metering change. The CUDA prover's batching budget, fractional-GKR work-buffer allocation, kernels, transcript/proof logic, and GKR scheduling stay on their existing paths. The selected CUDA strategy is only passed into `ProvingMemoryConfig`.

For non-default precompute-M tuning env vars, metering stays conservative because tuned thresholds can make non-last rounds fall back to fold-eval while precompute-M is globally enabled. That conservative tuned-env estimate does not change the prover allocation path.

The PR also keeps the more precise WHIR first-opening overlap model in `memory_metering.rs`, and adds serde compatibility for the new private metering field.

To cover CUDA resident memory that is not represented by per-segment trace/protocol counts, the CUDA backend now reports raw retained proving-key allocation sizes. `ProvingMemoryConfig` applies CUDA allocation rounding and incorporates the retained total into the segment estimate. This keeps the dependency direction as metering depending on backend allocation shape, not the prover depending on memory-metering config. The retained allocation walk counts retained PCS/tree/rule buffers and does not count the preprocessed trace buffer separately, since opened trace memory is already represented by the segment main-memory estimate.

## Validation

Local:

- `cargo +nightly fmt -- --check`
- `git diff --check`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo check -p openvm-stark-backend`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo test -p openvm-stark-backend memory_metering --features test-utils`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo clippy -p openvm-stark-backend --all-targets --tests -- -D warnings`

CUDA backend:

- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo check -p openvm-cuda-backend`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo clippy -p openvm-cuda-backend --all-targets --tests -- -D warnings`

OpenVM integration check with the matching metering call-site:

- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo check -p openvm-circuit`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo clippy -p openvm-circuit --all-targets --tests -- -D warnings`

openvm-eth, `execute-metered`, block `23992138`:

- CUDA backend enabled: yes
- completed: yes
- segments: `88`
- memory utilization over 88 segments: min `61.65%`, avg `78.25%`, max `97.87%`

openvm-eth, `prove-app`, `--segment-max-memory 16106127360`, with the matching OpenVM metering integration:

| Block | Segments | Max metered segment | Max CUDA-tracked peak | Peak minus max segment |
|---|---:|---:|---:|---:|
| `23992138` | `88` | `16,106,105,376` bytes | `15,391,912,044` bytes | `-714,193,332` bytes |
| `25399321` | `163` | `16,102,877,728` bytes | `15,822,868,764` bytes | `-280,008,964` bytes |

The maximum metered segment remains at or below the configured 15 GiB cap. The observed CUDA-tracked prover peak is below the corresponding segment estimate for both the default representative block and the heavier block.

The peak in both prove runs occurs during `prover.logup_zerocheck_prover:after_fractional_sumcheck_gkr`, so the remaining high-water mark is from the expected fractional-GKR proving phase plus resident main/proving-key state, not from metered execution allocating GPU buffers.

This recovers the reported regression from about `115` segments back to `88` for the representative default block while preserving the safer retained-memory accounting from #367.

### Benchmark runs

[block 23992138](https://github.com/axiom-crypto/openvm-eth/actions/runs/28465393386)
[block 25399321](https://github.com/axiom-crypto/openvm-eth/actions/runs/28465411633)

### DIFF
--- a/crates/cuda-backend/src/device.rs
+++ b/crates/cuda-backend/src/device.rs
@@ -1,5 +1,7 @@
 use getset::{CopyGetters, Getters, MutGetters};
-use openvm_cuda_common::{common::get_device, stream::GpuDeviceCtx};
+use openvm_cuda_common::{
+    common::get_device, memory_manager::vpmm_page_size_bytes, stream::GpuDeviceCtx,
+};
 use openvm_stark_backend::{
     memory_metering::ProvingMemoryConfig, StarkProtocolConfig, SystemParams,
 };
@@ -40,6 +42,10 @@ impl GpuProverConfig {
         let memory_config =
             ProvingMemoryConfig::from_protocol_config(config, self.cache_rs_code_matrix)
                 .with_cuda_backend(self.cache_stacked_matrix);
+
+        let memory_config = vpmm_page_size_bytes().map_or(memory_config, |page_size| {
+            memory_config.with_large_allocation_granularity_bytes(page_size)
+        });
 
         memory_config.with_fractional_gkr_work_buffer_strategy(
             fractional_gkr_work_buffer_metering_strategy(),
--- a/crates/cuda-backend/src/device.rs
+++ b/crates/cuda-backend/src/device.rs
@@ -4,9 +4,12 @@ use openvm_stark_backend::{
     memory_metering::ProvingMemoryConfig, StarkProtocolConfig, SystemParams,
 };
 
-use crate::cuda::{
-    batch_ntt_small::{ensure_device_ntt_twiddles_initialized, validate_gpu_l_skip},
-    device_info::get_sm_count,
+use crate::{
+    cuda::{
+        batch_ntt_small::{ensure_device_ntt_twiddles_initialized, validate_gpu_l_skip},
+        device_info::get_sm_count,
+    },
+    logup_zerocheck::fractional_gkr_precompute_m_enabled,
 };
 
 /// Pure device configuration — no stream, no CUDA runtime state.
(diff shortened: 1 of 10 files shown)


zlangley (approved):

stephenh-axiom-xyz:
Link the run please

shuklaayush:
closing for now since it's still not accurate and this is getting complicated

PR #371, opened 2026-07-01
chore(docs): bump version to `v2.0.0`

shuklaayush:
towards int-7836


CHANGELOG.md
shuklaayush, line 18: maybe we should add more here

stephenh-axiom-xyz (approved):

PR #375, opened 2026-07-09
fix: account for missing GPU allocations in proving memory estimates

yi-sun:
# Fix GPU proving memory estimates underestimating observed peaks (INT-8600)

## Overview

The metered-execution memory estimate (`memory_metering.rs`) was sometimes **below** the actual peak GPU memory observed during proving — up to **+3.2%** measured on the captured reth-block profile (e.g. estimate 5008 MiB vs measured peak 5167 MiB). On a tightly budgeted GPU this is an OOM risk: OpenVM uses `estimate().total` against `segmentation_max_memory` to decide segment boundaries.

Root cause (validated empirically on an RTX Pro 6000 and by a per-phase allocation audit of the CUDA prover): the true proving peak lands at the **WHIR round-0 query step**, where the prover *recomputes the full RS code matrix* (production runs with `cache_rs_code_matrix = false`) while the traces, the commitment Merkle digest layers, and the round-0 folded codeword + its tree are all still resident (`whir.rs:396-466`). The model priced that phase as bare `max(rs_code_matrix, …)`. Several smaller buffers were also unpriced.

There are no AIR or constraint changes in this PR; it only touches the host-side memory model, memory-manager instrumentation, and adds a validation benchmark.

## `crates/stark-backend` — memory model fix (reviewable on its own)

`estimate()` changes from

```text
total = main + max(rs_code_matrix, main_secondary, interaction)
```

to

```text
total = main + stacked_matrix_if_cached
      + max( rs_code_matrix + whir_overhead,   # opening phase: RS + WHIR working set coexist
             batch_secondary,                  # batch-constraint phase, explicit scratch
             interaction )                     # fractional-GKR phase
```

Term by term:

- **`whir_overhead_memory_bytes()`** (new): commitment-tree digest layers (`2·digest·C/2^k_whir`, still open for query proofs), round-0 folded codeword (`D_EF·F·C/2`), its tree, plus a fixed 64 MiB scratch term (`C = 2^(log_stacked_height + log_blowup)`). Added **on top of** `rs_code_matrix`, not maxed — measured at a stable ~140 MiB at 2^22 stacked height.
- **`rs_code_matrix_memory_bytes()`**: allows one extra stacked column, since stacking rounds the used width up to whole columns of height `2^(l_skip+n_stack)`.
- **`interaction_memory_bytes_without_overhead()`**: now mirrors the actual fractional-GKR work-buffer sizing, including its **2^22-entry floor** (`GKR_WINDOW_DEFAULT_MIN_N` in `fractional.rs`) which the old `logical_len/16` weight missed for small interaction counts (~90 MiB); fixed overhead raised 2 MiB → 64 MiB.
- **Batch-constraint term** (new): with `zerocheck_save_memory` the batch-MLE scratch is budgeted inside the GKR peak, so `max(main_secondary, gkr) + 192 MiB` (the constant covers eq segment trees, folded selectors, logup monomial tables — measured ~90 MiB). Without it (`log_blowup > 1`), the code greedily packs scratch up to `max(gkr_peak, 6 GiB)` **in addition to** `mat_evals` (no subtraction in that path), so the term becomes `main_secondary + max(gkr, BATCH_MLE_MEMORY_FLOOR) + 192 MiB`.
- **`cache_stacked_matrix`**: keeps a resident stacked-matrix copy for the whole proof; now added to `total`.

`ProvingMemoryConfig` gains `log_stacked_height`, `k_whir`, `digest_size`, `cache_stacked_matrix`, `zerocheck_save_memory` (defaults preserve the production coupling `zerocheck_save_memory = (log_blowup == 1)`). `ProvingMemoryEstimate` gains a `whir_overhead` component field. Constants that mirror CUDA-side sizing are documented as such (`GKR_MIN_WORK_BUFFER_LEN`, `BATCH_MLE_MEMORY_FLOOR`).

## `crates/cuda-backend` — config plumbing (2 lines)

`GpuProverConfig::proving_memory_config()` forwards the actual `cache_stacked_matrix` / `zerocheck_save_memory` flags into `ProvingMemoryConfig` so the estimate matches the prover configuration in use.

## `crates/cuda-common` — peak instrumentation

The prover's phase-level `MemTracker::reset_peak()` calls clobber `max_used_size`, so a whole-proof peak could not be read externally. Adds a `session_peak` counter that only `reset_session_peak()` resets, exposed via `tracked_memory_stats()`. No change to allocation behavior (two integer compares under the existing lock).

## `benchmarks/synthetic` — validation harness

New `mem_meter_runner` binary: replays captured segment profiles through `Coordinator::prove` on the GPU, replicates OpenVM's metered cell counting (`padded_height·total_width` split by `need_rot`, `padded_height·num_interactions`), and reports estimate-vs-measured peak with per-phase attribution from the `gpu_mem.*` gauges. Skips segments whose constraint degree exceeds the params limit; streams results as JSONL.

## Validation (RTX Pro 6000, reth-block profile, `app_params_with_100_bits_security(22)`)

| Configuration | Segments | Underestimates | Worst peak/estimate |
|---|---|---|---|
| **Before** (default flags) | 30 | 11 | **1.032** |
| After — default, full profile | 149 | 0 | 0.994 |
| After — `cache_rs_code_matrix = true` | 22 | 0 | 0.990 |
| After — `zerocheck_save_memory = false` | 44 | 0 | 0.994 |
| After — stacked height 2^20 | 48 | 0 | 0.992 |

Worst-case remaining headroom is 0.6%; conservatism is bounded at ~5–7% in the production metering regime. Estimates in that regime grow by ~200–400 MiB (~3–6%).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,9 +45,9 @@ Non-default CUDA support members (require GPU/CUDA toolchain — don't add to de
 
 Non-default benchmark workspace members:
 - **`openvm-benchmarks-fields`** (`benchmarks/fields`): CUDA field-extension and Poseidon2 benchmarks and verification tests.
-- **`openvm-benchmark-synthetic`** (`benchmarks/synthetic`): Synthetic AIR benchmark suite, including the champ-vs-candidate profile replay workflow documented in its README.
+- **`openvm-benchmark-synthetic`** (`benchmarks/synthetic`): Synthetic AIR benchmark suite, including the champ-vs-candidate profile replay workflow and the `mem_meter_runner` memory-metering validation runner documented in its README.
 
-When changing CUDA fractional-GKR buffer layout, scheduling, or scratch allocations, update the interaction memory estimate in `crates/stark-backend/src/memory_metering.rs` and the accounting in `docs/cuda-backend/gkr-prover.md`.
+When changing GPU buffer layout, scheduling, or scratch allocations in any proving phase — fractional-GKR, batch-constraint/batch-MLE, stacking/RS encoding, or WHIR opening — update the proving memory model in `crates/stark-backend/src/memory_metering.rs` (see `ProvingMemoryConfig::estimate` and the mirrored constants documented there), and validate against measured peaks with `mem_meter_runner`. For GKR changes, also update the accounting in `docs/cuda-backend/gkr-prover.md`.
 
 Plonky3 crates are exact-version crates.io dependencies pinned in the workspace `Cargo.toml`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Fixed
+- Proving memory estimates (`memory_metering`) no longer under-estimate the observed peak GPU memory: the estimate now accounts for the WHIR opening working set (commitment Merkle layers, round-0 codeword and tree, scratch) that coexists with the Reed-Solomon code matrix, stacking width round-up in the RS code matrix, the fractional-GKR work-buffer floor, batch-constraint scratch (additive on `main_secondary` when `zerocheck_save_memory` is off), and the resident stacked matrix when `cache_stacked_matrix` is on.
+
+### Added
+- `tracked_memory_stats`/`reset_session_peak` in `openvm-cuda-common` for peak GPU memory measurement that survives phase-level peak resets.
+- `mem_meter_runner` benchmark binary that replays captured segment profiles on the GPU and compares the memory-metering estimate against the measured peak.
+
 ## v2.0.0 (2026-07-06)
 
 ### Added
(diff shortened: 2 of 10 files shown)


yi-sun:
## Benchmark: standard reth block vs baseline (`origin/main`)

Setup: `synthetic_runner` (identical source on both revisions) replaying the reth-block-23992138 profile, `--sample-frac 0.2 --seed 7 --max-log-height 22` (30 degree≤3 segments, identical sampling), single Modal RTX Pro 6000 sandbox, interleaved **A/B/B/A** run order, two runs per binary.

| Metric (total prove over 30 segments) | baseline | this PR | delta |
|---|---|---|---|
| Mean of 2 runs | 9875 ms | 9980 ms | +1.07% |
| Per-segment median delta | — | — | **−0.10%** |
| Excluding first segment (process warmup) | 9060 ms | 9000 ms | −0.66% |

Run-to-run noise within the *same* binary is far larger than any A/B difference (the two baseline runs differ by −5.3% total; the 5 largest segments swing 13–29% across runs in both binaries). 25/30 segments are within ±0.6%.

Conclusion: **no measurable proving-performance change**, as expected — the only runtime-relevant diff is two integer comparisons under the existing memory-manager lock (`session_peak` tracking); the estimate-model changes are host-side arithmetic outside the proving path.

PR #377, opened 2026-07-10
feat(cuda-common): static GPU memory planner with allocation tracing

yi-sun:
## Motivation

GPU proving allocates buffers with heavily skewed sizes — from single-element sums to multi-GiB RS-code matrices — through the dynamic memory manager (VPMM). VPMM fights fragmentation at run time by remapping pages, which makes peak usage an emergent property of allocator state and makes pointer values nondeterministic across runs. For workloads like OpenVM, once trace heights are known every downstream allocation size and lifetime is a deterministic function of the proving key and config, so GPU memory can instead be planned statically.

This PR adds a static memory planner to `openvm-cuda-common`, the allocation tracing/analysis tooling used to evaluate it on real workloads, and a record-and-replay integration that runs existing provers on statically planned memory with no prover code changes. Design doc: `docs/static_alloc_spec.md`.

## Design

Explicit planning happens in two stages:

1. **Setup** — each phase declares its allocations through an `AllocBuilder`. `alloc_fake::<T>(len)` returns a `DeviceBufFake<T>`, an RAII handle owning no memory whose Rust lifetime *is* the declared lifetime of the future buffer; the builder's logical clock records every handle's birth/death interval, so ordinary ownership rules build the interference graph. `idx()` yields a small `Copy` handle (`AllocIdx<T>`).
2. **Run** — `build_on(&GpuDeviceCtx)` packs all declarations into one workspace (greedy best-fit-decreasing over the interference graph: deterministic, 256-byte aligned, lowest-offset first fit), allocates it on the given stream, and returns a `StaticAllocator` bound to that stream. `get(&idx)` returns a `StaticDeviceBuffer<T>` — an RAII view at the planned offset that derefs to `DeviceBuffer<T>`, so kernels and D2H copies work unchanged (H2D goes through an inherent `copy_from`; exposing `DerefMut` would let safe code `mem::replace` the non-owning view out of the guard and free workspace-interior memory). Pointers are stable across acquire/release cycles and proof rounds — the prerequisite for CUDA graph capture.

Safety nets: use-after-lifetime detection while conflicting guards are held (error or warn-and-allocate policy); planned reuse relies on declared-lifetime issue order on the bound stream, the same discipline as any explicitly allocated `DeviceBuffer`; `touchemall` poisons planned regions on each acquisition.

## Tracing, analysis, and replay integration

- `VPMM_TRACE=<path>` records every `d_malloc_on`/`d_free` (requested size, stream, timestamp, measured call latency) plus `MemTracker` phase markers.
- `AllocBuilder::plan_stats()` runs the packing pass without allocating; `benches/trace_replay.rs` and `benches/trace_phases.rs` replay traces offline (whole-trace plan vs dynamic peak; per-phase and per-proof-window breakdowns).
- `VPMM_REPLAY=<trace>` integrates the planner into the memory manager itself: at startup it plans all traced allocations of at least `VPMM_REPLAY_MIN` bytes (default 16 MiB), lazily allocates one workspace through the pool, and serves matching allocations at their planned offsets; frees of served pointers are bookkeeping only. Correctness does not depend on an exact replay: an allocation is served only when its size matches the next planned event and its planned range overlaps no live served allocation; anything else (including a second stream — replay is single-stream by design) falls back to the dynamic allocator with counters logged. The workspace base is rejected in `d_free` so a double-free of the offset-0 served pointer cannot release the workspace to the pool.

## Measurements (RTX Pro 6000, openvm-eth mainnet block 23992138, this codebase / v2.0.0 lineage)

Packing — the planner hits the peak-live lower bound exactly (efficiency 1.000) on every recorded trace; planning takes seconds for 10k–60k allocations:

| workload | allocs planned | static workspace | dynamic tracked peak | ws/peak |
|---|---|---|---|---|
| prove-app (84 segments) | 9,815 (≥16 MiB) | 14.0 GiB | 14.1 GiB | 98.8% |
| prove-stark (full aggregation) | 11,622 (≥16 MiB) | 14.0 GiB | 14.2 GiB | 98.3% |

End-to-end A/B with `VPMM_REPLAY` (real proving on planned memory; 9,815/9,815 allocations served, zero fallbacks across 9 runs): the allocation-heavy zerocheck/logup phase improves consistently (27.28 s → 26.50 s summed per proof, ±0.36 s over 5 paired runs), matching the ~0.8 s of dynamic-pool CPU it removes; total `app_prove` time is unchanged within noise (±3–6 s on ~176 s). The dynamic allocator is not a throughput bottleneck on this workload — the planner's value is the deterministic pre-known memory budget, allocator-independent peak behavior, and stable pointers for future CUDA graph capture of the segment loop.

Older-generation (v1.4.0) traces show the same packing behavior (12.5 GiB / 38.3 GiB workspaces at 99.0–99.5% of the dynamic peak).

## Limitations / follow-ups

- Explicit adoption inside the prover (declaring phase buffers through `AllocBuilder` rather than replaying a trace) is future work; heights are only known at prove time, so plans would be built or memoized per height shape after trace generation.
- The packer does not exploit cross-stream event dependencies (different stream keys never share memory).
- CUDA graph capture of the segment-proving loop on top of the stable-pointer property is the natural next step.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Added
+- (CUDA common) Static memory planner ([spec](./docs/static_alloc_spec.md)): declare allocations and their lifetimes up front via `AllocBuilder`/`DeviceBufFake` RAII handles, pack them into a single fixed workspace with an interference-graph packing pass, and serve fixed-offset buffers through `StaticAllocator::get` with run-time detection of use-after-lifetime. Includes allocation tracing (`VPMM_TRACE`), offline plan analysis tools, and a record-and-replay mode (`VPMM_REPLAY`) that serves a recorded workload's large allocations from a statically planned workspace with no prover code changes.
+
 ## v2.0.0 (2026-07-06)
 
 ### Added
--- a/crates/cuda-common/Cargo.toml
+++ b/crates/cuda-common/Cargo.toml
@@ -30,3 +30,15 @@ touchemall = []
 # operations (fill_zero, etc.) use the same stream. Useful for catching
 # accidental cross-stream misuse during development.
 debug-cuda-stream = []
+
+[[bench]]
+name = "static_alloc"
+harness = false
+
+[[bench]]
+name = "trace_replay"
+harness = false
+
+[[bench]]
+name = "trace_phases"
+harness = false
--- a/crates/cuda-common/benches/static_alloc.rs
+++ b/crates/cuda-common/benches/static_alloc.rs
@@ -1,0 +1,444 @@
+//! Static vs dynamic allocation benchmark.
+//!
+//! Replays a STARK-prover-like allocation schedule (skewed sizes, phase
+//! structure: commit -> perm -> quotient -> opening/FRI) twice:
+//!   1. dynamically, through `DeviceBuffer::with_capacity`/drop (VPMM pool);
+//!   2. statically, planning the same schedule with `AllocBuilder` and then serving every buffer
+//!      from the packed workspace.
+//!
+//! Reports peak GPU memory, wall time per round, and cumulative CPU time
+//! spent inside alloc/free (resp. get/drop) calls.
+//!
+//! Env knobs: `BENCH_NUM_AIRS` (default 40), `BENCH_LOG_MAX` (max log trace
+//! height, default 19), `BENCH_ROUNDS` (measured rounds, default 3),
+//! `BENCH_TOUCH` (memset every acquired buffer, default 1).
+//!
+//! Run: `cargo bench -p openvm-cuda-common --bench static_alloc`
+
+use std::time::{Duration, Instant};
+
+use bytesize::ByteSize;
+use openvm_cuda_common::{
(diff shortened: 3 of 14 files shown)


yi-sun:
CI benchmark on openvm-eth `develop-v2.1.0` (block 24001988, `prove-stark`, g7e.2xlarge, openvm pinned to `0b32f5dc`), base vs this PR — regression check for the default path (tracing/replay are env-gated and off by default):

|---|---|---|---:|---:|---:|
| base (`develop-v2.1.0`) | `7a425b64` | [30056302300](https://github.com/axiom-crypto/openvm-eth/actions/runs/30056302300) | 48.63 s | 67.85 s | 15.9 GiB |
| this PR | `2f16e976` | [30062463725](https://github.com/axiom-crypto/openvm-eth/actions/runs/30062463725) | 48.75 s (+0.23%) | 68.38 s (+0.78%) | 15.8 GiB |

**No regression**: every prover phase is within ±0.8% with no systematic direction (GKR −0.1%, round-0 +0.3%, WHIR +0.1%, commit +0.3%) — single-run noise, well below the deltas the same setup resolved for #380 (−3.15%) and #381 (−1.66%).

Metric is summed `stark_prove_excluding_trace_time_ms` over all proofs (app + leaf + internal); single CI run per config.

PR #379, opened 2026-07-10
perf(cuda): use warp shuffles for fractional bit reversal

yi-sun:
## Summary

- generalize the fused fractional bit-reversal and K=2 tree-build kernel to support Z=4 tiles
- replace the Z=4 shared-memory transpose with a two-stage four-lane register-shuffle butterfly
- make the shuffle path the default while retaining the original Z=8 and shared-memory Z=4 paths as rollback options
- add dense and virtual-padding proof-equivalence coverage and document the launch modes

Addresses INT-7891.

## Motivation

Profiling showed that the fused bit-reversal/tree-build kernel was the largest fractional-sumcheck component. Its Z=8 transpose used 64 KiB of dynamic shared memory per block and synchronized through shared memory while sustaining low warps in flight.

The Z=4 shuffle path removes dynamic shared memory and transpose barriers. It uses more registers than shared-memory Z=4, but produced no local-memory spills and was substantially faster on the target Blackwell GPU.

## Performance

Standard block 23992138, 78 cached segments, RTX Pro 6000, sequential main-to-candidate runs in the same sandbox:

| Metric | main | candidate | Change |
| --- | ---: | ---: | ---: |
| Fractional sumcheck | 10,369 ms | 8,575 ms | **-17.30%** |
| LogUp GKR | 10,458 ms | 8,669 ms | -17.11% |
| Zerocheck + LogUp GPU | 24,414 ms | 22,698 ms | -7.03% |
| Wall time | 421.9 s | 413.8 s | -1.92% |
| Peak VRAM | 16,157 MiB | 16,157 MiB | unchanged |

The candidate used the new default path without optimization environment variables.

Isolated shared-Z4 to shuffle-Z4 fractional-sumcheck results:

| Size | Shared Z4 | Shuffle Z4 | Change |
| --- | ---: | ---: | ---: |
| n=24 | 16.362 ms | 15.353 ms | -6.16% |
| n=26 | 32.335 ms | 29.198 ms | -9.70% |
| n=28 | 90.323 ms | 76.680 ms | -15.10% |

## Validation

- `cargo +nightly fmt --all -- --check`
- `CXX=g++-12 CUDA_ARCH=89 cargo check -p openvm-cuda-backend --tests`
- `CXX=g++-12 CUDA_ARCH=89 cargo clippy -p openvm-cuda-backend --tests -- -D warnings`
- GPU proof equivalence across Z8, shared Z4, and shuffle Z4 for dense and virtually padded inputs
- clean release build and end-to-end standard-block prove

## Rollback controls

- `SWIRL_CUDA_GKR_BITREV_Z_COUNT=8` restores the original Z=8 shared-memory kernel
- `SWIRL_CUDA_GKR_BITREV_Z4_SHUFFLE=0` selects the shared-memory Z=4 transpose

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+- (CUDA backend) Use a Z=4 register-shuffle transpose by default for the fused fractional-sumcheck bit-reversal and first two tree layers. Set `SWIRL_CUDA_GKR_BITREV_Z_COUNT=8` to restore the original Z=8 kernel or `SWIRL_CUDA_GKR_BITREV_Z4_SHUFFLE=0` to select the shared-memory Z=4 path.
+
 ## v2.0.0 (2026-07-06)
 
 ### Added
--- a/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
+++ b/crates/cuda-backend/cuda/src/logup_zerocheck/gkr.cu
@@ -1738,31 +1738,33 @@ T bit_rev(T i, unsigned int nbits)
 //
 // DERIVED FROM: bit_rev_permutation_z<T, Z_COUNT> in supra/ntt_bitrev.cu.
 // Diff vs the original template kernel:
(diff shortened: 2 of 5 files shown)


yi-sun:
This ended up being bad on CI benchmarks, closing.

PR #380, opened 2026-07-10
perf(cuda-backend): eliminate host round trips in sumcheck phases

yi-sun:
## Summary

Kernel-level profiling of the reth workload (CUPTI trace of the `benchmarks/synthetic` app-segment replay on RTX Pro 6000) showed the prover losing ~19% of prove wall time to host↔device round trips concentrated in the sumcheck phases: `mle_rounds` ran at 39% GPU utilization, round 0 serialized the GPU on per-AIR readbacks and host iDFTs, the stacked-reduction rounds issued 25+ small kernel launches each, and the logup-GKR/WHIR/stacked-reduction rounds each paid a pageable-copy readback per transcript round.

This PR eliminates those round trips phase by phase (one commit each):

1. **Batch MLE sumcheck rounds** (`logup_zerocheck`): new `RoundStager` packs all small per-round parameter tables (per-trace pointer tables, batch contexts) into one host staging buffer uploaded with a single H2D copy into a persistent device arena (was ~60 tiny pageable uploads per round). All monomial batch kernels of a round now launch back-to-back with reads deferred to the end of the round (was 5–6 blocking reads interleaved with launches). Per-trace interpolation buffers are persistent across rounds; trace+selector folding is one `batch_fold_mle` launch; `into_column_openings` gathers with async D2D copies and a single D2H read. DAG (rules) paths keep launch-then-read per memory batch so intermediate-buffer memory stays bounded as before.
2. **Pinned-memory readbacks** (`cuda-common::PinnedBuffer` + `MemCopyD2H::to_pinned_on`): per-round transcript readbacks in GKR, WHIR, and stacked reduction use page-locked memory (direct DMA, no pageable staging); the two per-round stacked-reduction scalar reads now share one stream sync.
3. **Batched stacked-reduction window launches**: new batched variants of the regular and degenerate MLE-round window kernels process all window groups of a round in one launch each (was one launch per unstacked matrix), preserving each group's former grid shape via a staged prefix-sum block table.
[...]

PR #381, opened 2026-07-10
perf(cuda-backend): fuse zeta and zero-extend passes into the NTT bit-reversal

yi-sun:
## Summary

Stacked on #380. Kernel profiling of the reth workload showed the standalone bit-reversal permutation (`bit_rev_permutation_z`) accounting for ~12% of prover kernel time: it runs as its own full read+write pass over the codeword buffer before every large NTT, in both the stacked commit and the WHIR rounds.

This PR removes the standalone passes by fusing the reordering into the adjacent passes, **without changing the committed codeword order, the proof, or the verifier**:

- **`bit_rev_zeta_fused`**: the RS encoding pipeline previously ran the subset-zeta stages (`x[j + 2^s] += x[j]` on the low `l_skip` index bits) and then the bit-reversal as two sweeps. The zeta chunks fit inside one row of the bit-reversal Z-tile, and under the tile's bit-reversed column storage, natural-index bit `s` maps to tile-column bit `LG_Z−1−s` — so the stages become in-tile column butterflies at strides `Z >> (s+1)`, applied while the tile is staged in shared memory. One pass instead of two; used by `rs_code_matrix` (commit and WHIR initial-round recompute).
- **`bit_rev_expand_pad`**: WHIR's per-round codeword previously did `batch_expand_pad` (zero-extend) and then a bit-reversal inside `batch_ntt`. The fused kernel writes `out[bit_rev(i)] = i < src_len ? in[i] : 0` directly through the Z-tile, again one pass instead of two.

Both kernels produce buffers bit-identical to the two-pass forms (the small-domain cases fall back to the original path). Verified end-to-end at production-like sizes (`uniform_runner` with stacked height 2^22 and 2^20, which exercises both fused kernels): proofs verify against the unchanged verifier.

### End-to-end (full reth `prove-stark`)

Measured on the real openvm-eth proof of block 23992138 (full recursive proof: app + leaf + internal + root, 224 proofs; openvm `develop-v2.1.0-rv64`), RTX Pro 6000. Steady-state (2 warmups), then main / #380 / this PR interleaved 3× in a Latin square (each config in each run-position once, so warmup drift and run-order bias cancel), mean of 3 reps; per-rep spread <0.3%. Metric is summed `stark_prove_excluding_trace_time_ms` over all 224 proofs. Incremental over #380 (this PR's base):

| metric | #380 | this PR | delta |
|---|---|---|---|
| prove (excl. trace) | 66.39 s | 65.10 s | **−1.9%** |
| total proof time | 117.11 s | 115.59 s | −1.3% |
| commit | 12.73 s | 12.25 s | −3.8% |
| openings (incl. WHIR) | 15.67 s | 15.04 s | −4.0% |

The gain shows up across the phases that RS-encode (commit and openings/WHIR), as expected. Cumulative #380 + this PR vs stark-backend main: **−5.5%** prove (excl. trace) / −3.5% total.

### Synthetic segment replay

`synthetic_runner` replaying app-proof segments of the captured block-23992138 profile (per-AIR `max_constraint_degree <= 3`; `--seed 42 --max-log-height 22`), shared Cargo.lock, interleaved champ (#380 tip) vs candidate:

| tier | #380 (ms prove) | this PR (ms prove) | delta |
|---|---|---|---|
| `--sample-frac 0.5` (75 segments, 2 reps interleaved) | 20249 / 20282 | 19792 / 19742 | **−2.46%** |
| `--sample-frac 0.1` (15 segments, 3 reps interleaved) | 4074 / 4035 / 4009 | 3960 / 3932 / 3936 | −2.39% |

## Test plan

- `uniform_runner` end-to-end prove+verify at stacked heights 2^22 and 2^20 (fused-kernel path): verifies
- `cargo nextest run -p openvm-cuda-backend -p openvm-cuda-common` on RTX Pro 6000: 131 passed, 2 skipped
- `cargo clippy --all-targets` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - (CUDA common) `DeviceBuffer::mut_slice(range)` returning a `DeviceBufferMutSlice<'_, T>` with a `copy_from_host` method for overwriting a subrange of a device buffer from a host slice on a caller-supplied stream.
 
 ### Changed
+- GPU prover: fused the subset-zeta and zero-extend passes of RS encoding into the bit-reversal permutation, removing the standalone bit-reversal sweeps before large NTTs (~2.5% lower proving time on the reth benchmark on RTX Pro 6000; committed codeword order and proofs unchanged).
 - GPU prover: batched the per-round parameter uploads, kernel launches, and result readbacks of the batch-constraint, stacked-reduction, and round-0 sumcheck phases, and moved per-round transcript readbacks to pinned memory; the interactions round-0 DAG is now cached in the proving key instead of rebuilt per proof. Measured ~2.4% lower proving time on the reth benchmark on RTX Pro 6000.
 
 ## v2.0.0 (2026-07-06)
--- a/crates/cuda-backend/cuda/supra/ntt_bitrev.cu
+++ b/crates/cuda-backend/cuda/supra/ntt_bitrev.cu
@@ -7,6 +7,8 @@
  * - 2025-08-13: Support multiple rows in bit_rev_permutation & bit_rev_permutation_z
  * - 2025-09-10: Add extern "C" launcher from sppark/ntt/ntt.cuh
  * - 2025-12-24: Template field type to support fr_t and bb31_4_t
+ * - 2026-07-10: Add fused subset-zeta + bit-reversal and expand-pad + bit-reversal
+ *               kernels (prover-side fusions; output identical to the two-pass forms)
  */
 
 #include <cstdint>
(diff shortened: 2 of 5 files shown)


yi-sun:
CI benchmark on openvm-eth `develop-v2.1.0` (block 24001988, `prove-stark`, g7e.2xlarge, openvm pinned to `0b32f5dc`), base (#380 tip) vs this PR:

|---|---|---|---:|---:|---:|
| base (#380 tip) | `44f9087d` | [30056324448](https://github.com/axiom-crypto/openvm-eth/actions/runs/30056324448) | 47.10 s | 66.56 s | 15.8 GiB |
| this PR | `e00aa870` | [30056346128](https://github.com/axiom-crypto/openvm-eth/actions/runs/30056346128) | 46.32 s (**−1.66%**) | 65.89 s (−1.00%) | 15.8 GiB |

The gain lands in the RS-encoding phases as intended: main trace commit −0.40 s, WHIR openings −0.46 s. In line with the −1.9% incremental in the description.

Metric is summed `stark_prove_excluding_trace_time_ms` over all proofs (app + leaf + internal); single CI run per config. Benchmarked revs differ from the current PR heads only by a rustfmt doc-comment rewrap in `pinned.rs`.

PR #382, opened 2026-07-10
perf(cuda-backend): prefetch the WHIR round-0 RS codeword on a low-priority stream

yi-sun:
## Summary

Stacked on #381. With `cache_rs_code_matrix = false` (the default), the WHIR initial round re-encodes the stacked matrix into the full RS codeword — trace stacking, per-chunk iNTT, subset-zeta/bit-reversal, and a full-size forward NTT — solely to gather the `num_queries * 2^k_whir` opened rows for the proof. This runs serially on the main stream inside the WHIR round. Caching the commit-time codeword instead (`cache_rs_code_matrix = true`) avoids the re-encode but keeps a full codeword resident from commit to WHIR, which is unacceptable on memory-constrained GPUs.

This PR launches the same re-encode ahead of time on a dedicated lowest-priority CUDA stream, right after the GKR fractional sumcheck (the phase with the largest transient memory) completes. The encode kernels yield to the main stream and soak the idle bubbles of the round-0 / batch-MLE / stacked-reduction phases, so when WHIR reaches the query openings the codeword is already (mostly) available: the main stream waits on the prefetch event and gathers. Proof bytes are identical — same kernels, same inputs, exact field arithmetic — and a new test asserts byte-equality of proofs with the prefetch on vs off.

Controlled by `GpuProverConfig::prefetch_rs_code_matrix` (default on); commits without a prefetched codeword (preprocessed/cached commits, or the flag off) fall back to the existing in-round recompute.

## Memory

Peak device memory is kept within the existing envelope:
- The codeword is resident only from post-GKR to the WHIR initial round, not for the whole proof as with `cache_rs_code_matrix`.
- The batch-MLE working buffers keep the `memory_limit_bytes` budget that keygen sized them to; the prefetched codeword is a separate pool allocation the allocator already tracks. (An earlier revision shrank this budget by the codeword size to "reserve" for it — that drove the round-0 batch-MLE below its working floor and deadlocked on the real e2e prover; see **Correctness**.)
- The codeword is **allocated on the main stream** while only the encode kernels run on the auxiliary stream. This matters for the VPMM: free regions are reusable same-stream without event gating, but cross-stream reuse requires completed free events and regions of different streams never coalesce — allocating multi-GB buffers on a second stream fragments the pool and triggers page-allocation/defrag churn (measured as a 28% regression before this fix). The auxiliary stream is ordered after the allocation point by an event, which also orders it after trace production.

## Correctness

The initial prefetch revision subtracted the resident codeword's size from the batch-MLE `memory_limit_bytes` after launching the re-encode, to keep peak demand flat. On the real end-to-end reth prover (`prove-stark`, whose app proofs run with `zerocheck_save_memory = true`) this drives the round-0 batch-MLE onto a degenerate low-memory path whose device-to-host readback **deadlocks** — the proof hangs indefinitely, busy-spinning in `cuMemcpyDtoHAsync` inside `sumcheck_uni_round0_polys`. `memory_limit_bytes` is the batch-MLE working-buffer hint sized by keygen, not a whole-proof accounting knob; the codeword is a separate pool allocation the allocator already tracks (and which returns a recoverable OOM rather than corrupting the batch budget). The subtraction (and the now-unused `codeword_bytes` helper) is removed, so the budget stays at the keygen value, matching the non-prefetch path. Proof output is byte-identical and verifies.

### Synthetic (app-proof segments)

Modal RTX Pro 6000 (Blackwell, 96 GB), `synthetic_runner` replaying app-proof segments of the captured reth block 23992138 profile (per-AIR `max_constraint_degree <= 3`, `--seed 42 --max-log-height 22`), shared Cargo.lock, VPMM pool 16 GiB, interleaved champ (#381 tip) vs candidate:

| tier | #381 (ms prove) | this PR (ms prove) | delta |
|---|---|---|---|
| `--sample-frac 0.5` (75 segments, 6 reps interleaved) | 20614 / 20501 / 20462 / 20611 / 20595 / 20485 (mean 20545) | 20371 / 20018 / 19955 / 20027 / 20033 / 19972 (mean 20063) | −2.35% |
| `--sample-frac 0.1` (15 segments, 3 reps interleaved) | 4244 / 4077 / 4074 | 3988 / 3970 / 3989 | −3.5% |

### End-to-end (openvm-eth, reth block 23992138)

The synthetic gain **does not reproduce on the full recursive proof**. Modal RTX Pro 6000 (Blackwell, 96 GB), `openvm-reth-benchmark --mode prove-stark` on the standard block (app + leaf + internal + root, 224 proofs), interleaved base (`perf/ntt-bitrev-fusion` tip = this PR's base) vs this PR HEAD, 2 warmups + 8 measured runs in balanced ABBA|BAAB order, shared Cargo.lock, VPMM pool 16 GiB. Metric is summed `stark_prove_excluding_trace_time_ms` over all proofs.

| metric | base (#381 tip) | this PR | delta |
|---|---|---|---|
| prove (excl. trace) | 67.22 / 65.61 / 65.76 / 65.69 (mean 66.07 s) | 66.53 / 66.06 / 66.30 / 65.67 (mean 66.14 s) | **+0.11%** |
| total proof time | mean 117.48 s | mean 118.69 s | +1.03% |
| WHIR openings | mean 11.48 s | mean 5.13 s | −55.3% |

The WHIR-openings span halves — the round-0 re-encode leaves WHIR's critical path as intended — but that does not turn into an end-to-end saving here: on the full recursive workload the sumcheck phases the encode overlaps are already GPU-saturated, so it competes for the same SMs rather than filling idle bubbles, and prove-excluding-trace time is flat (within run-to-run noise). Proof output is byte-identical (315287 bytes) and verifies in every run.

## Test plan

- `cargo nextest run -p openvm-cuda-backend -p openvm-cuda-common --test-threads=4` on RTX Pro 6000: 132 passed (including the new `test_rs_codeword_prefetch_proof_unchanged` byte-equality test), 2 skipped
- End-to-end `prove-stark` of reth block 23992138 with the prefetch active: completes without hang and the proof verifies (10/10 runs, byte-identical 315287-byte proof)
- `uniform_runner` end-to-end prove+verify at stacked heights 2^22 and 2^20 with the prefetch active: verifies
- `cargo clippy --all-targets -- -D warnings` clean, `cargo +nightly fmt -- --check` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - (CUDA common) `DeviceBuffer::mut_slice(range)` returning a `DeviceBufferMutSlice<'_, T>` with a `copy_from_host` method for overwriting a subrange of a device buffer from a host slice on a caller-supplied stream.
 
 ### Changed
+- GPU prover: when the RS codeword is not cached, the WHIR initial-round re-encode is now launched ahead of time on a low-priority auxiliary CUDA stream (right after the GKR fractional sumcheck), overlapping the batch-constraint and stacked-reduction sumcheck phases instead of running serially inside WHIR (~2.4% lower proving time on the reth benchmark on RTX Pro 6000). Controlled by `GpuProverConfig::prefetch_rs_code_matrix` (default on); proofs unchanged.
 - GPU prover: fused the subset-zeta and zero-extend passes of RS encoding into the bit-reversal permutation, removing the standalone bit-reversal sweeps before large NTTs (~2.5% lower proving time on the reth benchmark on RTX Pro 6000; committed codeword order and proofs unchanged).
 - GPU prover: batched the per-round parameter uploads, kernel launches, and result readbacks of the batch-constraint, stacked-reduction, and round-0 sumcheck phases, and moved per-round transcript readbacks to pinned memory; the interactions round-0 DAG is now cached in the proving key instead of rebuilt per proof. Measured ~2.4% lower proving time on the reth benchmark on RTX Pro 6000.
 
--- a/crates/cuda-backend/src/device.rs
+++ b/crates/cuda-backend/src/device.rs
@@ -1,12 +1,20 @@
+use std::sync::{Arc, Mutex};
+
 use getset::{CopyGetters, Getters, MutGetters};
-use openvm_cuda_common::{common::get_device, stream::GpuDeviceCtx};
+use openvm_cuda_common::{
+    common::get_device,
+    stream::{CudaStream, GpuDeviceCtx, StreamGuard},
+};
 use openvm_stark_backend::{
     memory_metering::ProvingMemoryConfig, StarkProtocolConfig, SystemParams,
 };
 
-use crate::cuda::{
-    batch_ntt_small::{ensure_device_ntt_twiddles_initialized, validate_gpu_l_skip},
-    device_info::get_sm_count,
+use crate::{
+    cuda::{
+        batch_ntt_small::{ensure_device_ntt_twiddles_initialized, validate_gpu_l_skip},
+        device_info::get_sm_count,
+    },
+    stacked_pcs::RsCodewordPrefetch,
 };
 
 /// Pure device configuration — no stream, no CUDA runtime state.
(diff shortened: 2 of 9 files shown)


yi-sun:
CI benchmark on openvm-eth `develop-v2.1.0` (block 24001988, `prove-stark`, g7e.2xlarge, openvm pinned to `0b32f5dc`), base (#381 tip) vs this PR:

|---|---|---|---:|---:|---:|
| base (#381 tip) | `e00aa870` | [30056346128](https://github.com/axiom-crypto/openvm-eth/actions/runs/30056346128) | 46.32 s | 65.89 s | 15.8 GiB |
| this PR | `209d9bbc` | [30056369475](https://github.com/axiom-crypto/openvm-eth/actions/runs/30056369475) | 45.95 s (**−0.79%**) | 65.30 s (−0.90%) | **20.8 GiB** |

The WHIR openings span drops 8.41 s → 3.23 s (−62%) while the `rap_constraints.round0` span absorbs +5.1 s as the prefetch competes for SMs underneath it — netting a small gain on this block (the description measured flat on block 23992138). Within single-run noise, so treat as ~neutral-to-slightly-positive.

Note the peak GPU memory increase: 15.8 → 20.8 GiB (+5 GiB for the codeword resident from post-GKR to the WHIR initial round) — relevant for memory-constrained GPUs.

Metric is summed `stark_prove_excluding_trace_time_ms` over all proofs (app + leaf + internal); single CI run per config. Benchmarked revs differ from the current PR heads only by a rustfmt doc-comment rewrap in `pinned.rs`.

PR #384, opened 2026-07-15
perf(cuda-common): pinned host-memory pool for H2D staging buffers

GunaDD:
Copies from pageable memory run at a fraction of PCIe bandwidth, but page-locking a buffer is itself expensive (~1-2 GB/s), and buffers are often provisioned at full capacity while only partially written, so neither registration nor re-zeroing may sit on the caller's critical path. 

Dropped buffers are therefore handed via [`give_back`] to a background cleaner thread which registers them once (`cudaHostRegister`), zeroes the prefix the previous owner wrote, and only then returns them to the pool. [`take`] hands out ready (registered, all-zero) buffers on a pool hit; on a miss it falls back to a fresh pageable allocation.

### DIFF
--- a/crates/cuda-common/src/lib.rs
+++ b/crates/cuda-common/src/lib.rs
@@ -3,4 +3,5 @@ pub mod copy;
 pub mod d_buffer;
 pub mod error;
 pub mod memory_manager;
+pub mod pinned;
 pub mod stream;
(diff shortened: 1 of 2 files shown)


gaxiom (approved):
Looks ok

PR #388, opened 2026-07-21
fix: pinned pool lifecycle

GunaDD:
Fixes the following issues: 

The current locked cuda-common pool still has several problems.
- It is process-global, but returned buffers do not carry their originating CUDA device or completion event.
- The cleaner is a new thread and calls cudaDeviceSynchronize() without selecting the originating device.
- In a process using GPU 0 and GPU 1, it could synchronize GPU 0 and then reuse a buffer still being read by GPU 1.
- The cleaner never shuts down or joins cleanly.
- If device synchronization fails, an already-registered pooled buffer can be dropped without cudaHostUnregister, while its pointer remains in the registered-pointer set.
- The pool retains an unlimited number of pinned buffers at every size class.

The robust design is for every returned buffer to carry its device and a CUDA event recorded after the last consuming copy. The cleaner waits for that exact event before zeroing or reusing it. Add an explicit shutdown/drain path and a global/per-size pinned-byte cap with eviction.

I tested on the PoC in `gunadi/pinned-module-poc` branch and it passed the PoC tests now.

### DIFF
--- a/crates/cuda-common/src/pinned.rs
+++ b/crates/cuda-common/src/pinned.rs
@@ -24,6 +24,9 @@
 //! [PinnedBuffer::record_last_use]; the cleaner then waits on exactly those
 //! events. Without a recorded event it falls back to synchronizing the device
 //! that was current when the guard dropped.
+//!
+//! The pool retains at most [set_max_pooled_bytes] bytes (default 4 GiB);
+//! past that, returning a buffer evicts pooled ones, largest first.
 
 use std::{
     collections::BTreeMap,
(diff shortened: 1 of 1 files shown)


crates/cuda-common/src/pinned.rs
```diff
@@ -9,42 +9,55 @@
...
     fn cudaHostRegister(ptr: *mut c_void, size: usize, flags: u32) -> i32;
     fn cudaHostUnregister(ptr: *mut c_void) -> i32;
 }
 
-/// A returned buffer together with its dirty-prefix length.
-type ReturnedBuffer = (Vec<u8>, usize);
+/// Registered memory is pinned for all CUDA contexts, so copies issued from
+/// any device get the fast path; without this flag, contexts other than the
+/// registering one (the cleaner's) would treat the buffer as pageable.
+const CUDA_HOST_REGISTER_PORTABLE: u32 = 0x1;
```
Aalanli, line 54: Does this make registering more expensive? If it does then this is not necessary as there is only one cuda context per process.

crates/cuda-common/src/pinned.rs
```diff
@@ -62,117 +75,364 @@ pub fn unregister_region(ptr: *mut u8) {
...
+/// prefix, and pools it. The buffer must be quiescent (no copies in flight).
+fn recycle(mut returned: Returned) {
+    if !returned.registered {
+        if !register_region(returned.data.as_mut_ptr(), returned.data.len()) {
+            return; // stays pageable; drop normally
+        }
+        returned.registered = true;
+    }
+    let dirty_len = returned.dirty_len.min(returned.data.len());
+    returned.data[..dirty_len].fill(0);
```
Aalanli, line 375: why do you need to zero buffers? This seems counter to malloc/free design.

crates/cuda-common/src/pinned.rs
```diff
@@ -9,42 +9,55 @@
...
+use crate::{
+    common::{get_device, set_device_by_id},
+    error::CudaError,
+    stream::{device_synchronize, CudaEvent, CudaStream},
+};
 
 #[link(name = "cudart")]
 extern "C" {
     fn cudaHostRegister(ptr: *mut c_void, size: usize, flags: u32) -> i32;
     fn cudaHostUnregister(ptr: *mut c_void) -> i32;
```
Aalanli, line 48: Instead of cudaHostRegister, to allocate pinned memory you could also use cudaMallocHost.

crates/cuda-common/src/pinned.rs
GunaDD: in the multi gpu setting could there be more than one cuda context per process ?

crates/cuda-common/src/pinned.rs
GunaDD: the pool isn't modeling the malloc/free design, it's modelling a fresh allocation from a zero-ed buffer. the buffer needs to be zeroed because of the invariant of the pool.

crates/cuda-common/src/pinned.rs
GunaDD: thanks! claude thinks in this case cudaHostRegister is a better fit

crates/cuda-common/src/pinned.rs
Aalanli: you could be right, I'm not sure how multigpu is architected here

crates/cuda-common/src/pinned.rs
```diff
@@ -62,112 +75,359 @@ pub fn unregister_region(ptr: *mut u8) {
...
+    /// Records the point on `stream` after which the buffer is no longer
+    /// read, i.e. right after enqueueing the last copy out of it there. The
+    /// cleaner then waits on exactly the recorded events — instead of a
+    /// whole-device synchronize — before reusing the buffer, which is also
+    /// what makes reuse safe when several devices are active.
+    ///
+    /// Call from the thread that enqueued the copies, with the same device
+    /// current (CUDA requires the event and stream to share a context). If
+    /// several streams consume the buffer, record on each of them.
+    pub fn record_last_use(&mut self, stream: &CudaStream) -> Result<(), CudaError> {
```
Aalanli, line 125: Yeah, this is a bit unfortunate. You should document that the caller **has to record this** before dropping. 
Otherwise the buffer could be reused without syncing the stream.

Aalanli:
Hi @GunaDD, overall the implementation seems correct to me. From my understanding, memory reuse comes from the best fit among the pool allocated? So if I allocate 1000, free it, then allocated 1 and 999, then a 999 buffer gets allocated on top of the 1000? 

Sorry I didn't point this out earlier, but I think this functionality could be better achieved by using a cuda memory pool, which tracks the stream for you and uses a better algorithm for memory reuse. Search `cuda memory pool host memory` and you'll get some results.

crates/cuda-common/src/pinned.rs
GunaDD: documented

GunaDD:
Hi @Aalanli, on the memory reuse question: it's not the best fit. From your example, the 1000 allocation becomes a 1024-byte buffer and that is what the pool holds after the free. The `take(1)` afterwards rounds to 1 and misses, so it gets a fresh allocation. The `take(999)` afterwards rounds to 1024 and hits, so it's true that the 999 lands on top of the 1000's allocation. 

On the cuda memory pool, I think it is out of the scope of this PR since this PR is supposed to be the fix to the current pinned memory module. Maybe we can do it in a follow up later ? Wdyt ?

Aalanli (approved):

PR #392, opened 2026-07-29
feat: degree-5 extension for baby bear field 

GunaDD:
Closes INT-8810


stephenh-axiom-xyz:
Let's not merge this into main? We will want to tag 2.0.2 soon, and this shouldn't be in that

GunaDD:
sounds good, lets wait until the right time

PR #395, opened 2026-08-21
feat(cuda-backend): express the logup-zerocheck phase as a graph-IR m…

gaxiom:
…odule

Mirror the existing fractional-sumcheck IR port for the logup-zerocheck phase. `logup_zerocheck_gpu_ir` emits stages C+D+E of `prove_zerocheck_and_logup_gpu` onto a single `GraphBuilder` as a full loop unroll, and `GraphCompiler::compile` produces a `GraphExe` from it.

Follows the conventions of `fractional_ir.rs`: `*_ir` inserter naming, one `insert_blackbox_kernel` per inserter, closures that only launch and `mem::forget` every `from_raw_parts` view, and in-file tests asserting on raw device bytes against the eager path.

Blackbox/Tier-1 throughout: no `ir::Module` DSL ports, and the eager path is untouched apart from visibility widening so the IR mirror can reuse its policy helpers. Enablement-only -- nothing is wired into the prove path, so `mod.rs` still calls the eager implementation.

Transcript seeding: the phase starts mid-protocol, so `DuplexSpongeGpu` gains a `snapshot()` export and `DuplexSpongeGpuIR` a `from_live()` constructor; the graph transcript can now continue an in-flight sponge instead of starting from a zeroed state.

Adds `examples/dump_ir_zerocheck_phase.rs` to build, dump and compile the phase graph, alongside the existing `dump_ir_*` programs.

Known gaps are marked `TODO(cc-ir)` with the ambiguity and the risk if the guess is wrong: the per-round host seam around `compute_batch_s_poly`, challenge-by-value kernels that need `_dev_challenge` siblings, round 0's host iDFT tail, multi-kernel launchers behind a single blackbox node, and host-assembled ctx structs.

Gates on GPU 7: cargo check --all-targets, +nightly fmt --check, clippy --all-targets --tests -D warnings, and nextest 155/155 passed.


Aalanli:
Hi @gaxiom , I pushed some stuff that should improve a lot of things. You can read 

notes.md for some lower-level gotchas: https://github.com/openvm-org/stark-backend/pull/391/changes#diff-61c8787a86901fe04a0e169dcd852444d8642bbfdbc9602a0e009a3e2ed11d59

and tools.md: https://github.com/openvm-org/stark-backend/pull/391/changes#diff-8b9000576e4fcea712a2ce22285d51c7267fda47a368bbd06e27c95c99b9167f

Feel free to refactor the compiler frontend to make building the graph easier to do.

crates/cuda-backend/src/logup_zerocheck/zerocheck_ir.rs
```diff
@@ -0,0 +1,4634 @@
...
+    b.push(data);
+    let (inputs, modifies, ptrs) = b.finish();
+    let reads = declared_reads(&inputs);
+    g.insert_blackbox_kernel(
+        "materialize_main_matrix_ptr",
+        inputs.into_iter(),
+        std::iter::empty(),
+        modifies.into_iter(),
+        move |inputs, _outputs, stream| unsafe {
+            materialize_main_matrix_ptr_raw(
```
Aalanli: I don't think this is safe, because you are assigning a pointer to the data to an output buffer.
This aliases the input pointer buffer lifetime to the lifetime of the output pointer of pointer buffer. The compiler doesn't know that the original lifetime could be extended and could reuse the existing buffer.

crates/cuda-backend/src/logup_zerocheck/zerocheck_ir.rs
```diff
@@ -0,0 +1,4634 @@
...
+//      (`fractional_ir.rs:922-955`), which is the better long-term answer.
+// RISK: as long as `xi` comes from the host (it does — GKR returns it to
+//      the host), this is exactly what the eager path does and is bit-exact.
+//      It blocks a fully device-resident Fiat-Shamir chain, nothing else.
+pub fn eq_hypercube_interleaved_stage_ext_ir(
+    g: &mut GraphBuilder,
+    input: BufId,
+    out: BufId,
+    x_i: EF,
+    step: u32,
```
Aalanli: see `fractional_sumcheck_ir.rs`, you can just promote each x_i to a BufId.

crates/cuda-backend/src/logup_zerocheck/zerocheck_ir.rs
```diff
@@ -0,0 +1,4634 @@
...
+                    sels_len: 3 << n_lift,
+                    main_ptrs_len: tp.mats.len() - usize::from(tp.has_preprocessed),
+                    public_len: tp.num_public_values,
+                },
+            );
+        } else {
+            // No producer otherwise — keep the registered output well-formed.
+            g.insert_memset(zc_out, 0);
+        }
+        g.register_output(zc_out);
```
Aalanli: Let's not `register_output` here. The output is positional and it gets hard to track otherwise. The IR builder should not `register_output`.

Aalanli:
Since there's no end to end working prototype yet, I'll mention a few high-level things:
1. It's better to modify the compiler frontend to make the IR building easier/less verbose, than introduce builder abstractions in the file. I think we can make the blackbox kernel construction easier and add type info to a wrapper of BufId, i.e. with a phantom type. 
2. I find with large-scale AI ports it's very important to have sub-parts done first and verified for correctness. I.e. for fractional_sumcheck I chose to port first the fold-eval strategy and after that worked, port the precompute-M strategy. I believe that's also the case here. I think it's also possible to move more of the strategy decision making to the compiler via the alternative graph saturation mechanism in fusion (eg. you have two possible computation graphs that compute the same thing, but with different memory and compute requirements, the compiler can decide which one based on the setup and constraints; maybe for batched high-throughput it's better to trade for less memory for more compute, and the opposite for low-latency).
3. I think it's also beneficial to maintain a porting guide (which I have in my branch). I like to write the porting guide myself, I get better results.

Aalanli:
I don't know what's your workflow, but I like to make a high-level plan, then ask AI to make a detailed plan which a checklist that I check into source. Then it should be clear for reviewers what stage the plan is at.

Aalanli:
Another thought is that we should first review the human-written plan and detailed plan first instead of the code. Then after pushing AI though review the code.

PR #396, opened 2026-08-26
feat(scheduler POC): resource-aware DAG scheduling engine and release-stream buffer ownership

gaxiom:
Adds `openvm-scheduler`, a dependency-free engine that runs independent nodes of an execution graph concurrently under an explicit resource budget, and fixes a stream-ordering defect in the CUDA allocator that concurrent proving exposes.

  ### `openvm-scheduler` (new crate)

  A ready-queue scheduler over a DAG of nodes, each declaring a resource profile, admitted against a budget. Additive: no existing crate changes behaviour, and nothing depends on it unless it opts in. The first consumer models execution and proving as a two-node graph so both a CPU stage and a GPU stage stay busy, with proving prioritised.  Node granularity is deliberately not fixed by the engine — a node can later be split into finer sub-nodes with their own dependencies and memory profiles without touching the engine.

  ### Release-stream buffer ownership

  With two proves resident on distinct streams, a buffer was freed via `cudaFreeAsync` on the stream that **allocated** it rather than the stream that last **read** it. A trace buffer could therefore be recycled while another stream was still reading it, producing well-formed but unverifiable proofs — the failure surfaces downstream as `LogupZerocheck: Fractional sumcheck: nonzero root sum`, at a rate around one run in three under concurrency, and never under serial proving.

  The allocation record now tracks the stream that must order release, with an explicit handoff at the ownership-transfer boundary. Rebind and free are serialised by the manager mutex, with sole ownership enforced rather than assumed. No device-wide barrier and no cross-prove synchronisation are introduced; measured overlap between two concurrent proves is unchanged.

  ### Scope

  This is a targeted ownership handoff for scheduler-owned trace buffers, **not** general reader tracking. Allocations read on a stream other than their allocator's, without an explicit handoff, still release on the allocating stream; the serial path is untouched and unaffected.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,13 +30,14 @@ cargo +nightly fmt -- --check  # check only
 
 ## Project Structure
 
-Rust workspace with 5 default-member crates:
+Rust workspace with 6 default-member crates:
 
 - **`openvm-stark-backend`** (`crates/stark-backend`): Core SWIRL proof system — prover, verifier, AIR builders, interactions, keygen, PCS. Re-exports Plonky3 crates.
 - **`openvm-codec-derive`** (`crates/stark-backend/codec-derive`): Proc macro for serialization codegen.
 - **`openvm-stark-sdk`** (`crates/stark-sdk`): Concrete configs (`BabyBearPoseidon2Config`, `BabyBearBn254Poseidon2Config`), tracing/metrics, benchmarks.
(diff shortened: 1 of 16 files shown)


gaxiom:
NOTE! All stream-related staff is temporary AI slop. I know how to do it cleaner but pushed for being able to CI on openvm-eth

